PR #107, opened 2015-09-30
carthage 1.0.0

vasarhelyia:
Step for setting up a project with [`Carthage`](https://github.com/Carthage/Carthage) dependency manager.


viktorbenei:
Thank you for the contribution! Should be available in the StepLib in a couple of minutes!

PR #169, opened 2015-11-18
AndroidSdkUpdate 1.0.0

guitcastro:
AndroidSdkUpdate step

### DIFF
--- a/steps/AndroidSdkUpdate/1.0.0/step.yml
+++ b/steps/AndroidSdkUpdate/1.0.0/step.yml
@@ -1,0 +1,74 @@
+title: Android SDK Update
+summary: "Need to install or update a packge from Android SDK Manager? Use this Step
+  to\ninstall and update them as you wish!   "
+description: 'Install the necessary sdk tools to compile your application  '
+website: https://github.com/guitcastro/steps-update-sdk
+source_code_url: https://github.com/guitcastro/steps-update-sdk
+support_url: https://github.com/guitcastro/steps-update-sdk/issues
+published_at: 2015-11-18T19:02:23.319877987-02:00
+source:
+  git: https://github.com/guitcastro/steps-update-sdk.git
+  commit: 99ae9775bf9300aa9f07a7ca02690c7ce752505a
+host_os_tags:
+- osx-10.10
+project_type_tags:
+- android
+type_tags:
+- android
+- build
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: false
+run_if: ""
+inputs:
+- opts:
+    description: Mark this if you want to update the Android SDK Build-tools.
+    is_dont_change_value: false
+    is_expand: false
+    is_required: true
+    summary: ""
+    title: Update Android SDK Build-tools
+    value_options:
+    - "on"
+    - "off"
+  tools: "on"
+- opts:
+    description: Select which version if Android SDK Platform-tools should be installed.
+    is_dont_change_value: false
+    is_expand: false
+    is_required: true
+    title: Android SDK Platform-tools version
+    value_options:
+    - stable
+    - preview
+    - both
+    - none
+  platform_tools: stable
+- build_tools: lasted
+  opts:
+    description: "List separated by comman all Build-tools versions that should be
+      installed. \nleave empty with don't want to install."
+    is_dont_change_value: false
+    is_expand: false
+    is_required: true
+    summary: Example:23.0.2,22.0.1
+    title: Android SDK Build-tools versions
+- opts:
+    description: "List separated by comman all platform versions that should be installed.
+      \nleave empty with don't want to install."
+    is_dont_change_value: false
+    is_expand: false
+    is_required: true
+    summary: 'Example: 23,22,17'
+    title: Android SDK Platform versions
+  platform: "23"
+- opts:
+    description: |-
+      List separated by comman of all system images that should be installed.
+      Use `android list --all -e` to get see the name of all SystemImages
+    is_dont_change_value: false
+    is_expand: false
+    is_required: true
+    summary: 'Example: sys-img-armeabi-v7a-android-tv-23, sys-img-armeabi-v7a-android-22'
+    title: SystemImages
+  system_images: none


viktorbenei:
First of all thank you for the PR & contribution, we really value it a **lot**! There are just a couple of things which should be changed, and other which we thought would be a nice addition / revision.

Suggested changes:

* you should change the step ID from `AndroidSdkUpdate` to a lowercase ID, like `android-sdk-update`, as that's how the other step IDs look like, and for someone who works with the `bitrise.yml` directly following another ID naming convention will be confusing
* there's a typo here in the default value: https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L42 - `lasted` should be `latest`
* your README is for the old StepLib, you should use a newer text like the one here: https://github.com/bitrise-io/bitrise/blob/master/_step_template/README.md - especially because the Step sharing works differently now than it did for the old StepLib
* If possible you should check the parameters individually, whether a given `filter` is already installed. The problem with not checking it is that even if it's already installed, Android will try to re-install it, which might take a long time. There's an additional issue related to Docker: if a tool is already installed Android will try to create a backup copy with a symlink type which is not supported by `docker`, and so it'll always fail.


Other notes & comments:

* you can remove the `bash_utils` folder as you don't use it
* what does the `tools ` input do? ( https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L23 ) turning it off could be the same as not providing any `build_tools`, just change `is_required: true` to `false` ( https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L49 ) and handle this case in the code, and you should be able to remove the `tools` input. Less input is always preferred :)
* similarly you could change the `system_images: "none"` input to `is_required: false` and handle the empty input the way you handle `"none"` right now. Less typing, smaller chance for typos ;)

Most of these are just minor things, but please consider all of the points, as it would improve the experience of using your step!

Happy building! 😊

viktorbenei:
Ohh, there's another minor typo here: https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L37 - `version if Android` should be `version of Android`

guitcastro:
Hi viktorbenei,

Thanks for the feedback! I made all the possible changes that you suggested.
Here are some notes about the ones I couldn't do:

*"If possible you should check the parameters individually, whether a given filter is already installed. The problem with not checking it is that even if it's already installed, Android will try to re-install it, which might take a long time. There's an additional issue related to Docker: if a tool is already installed Android will try to create a backup copy with a symlink type which is not supported by docker, and so it'll always fail."*

* I agree with you, check if the package is already installed will be awesome, but I couldn't find out how to check this.

*"what does the tools input do?"*

* The tools input is to update the Android SDK Manager itself. Not related to any version of the Android, the executables locate in the **tools** folder in the SDK.

viktorbenei:
hi @guitcastro - just a heads up, we didn't forget about your PR but we're searching for a solution to either check whether a tool is already installed, and/or to solve the Docker file system issue.

So far it seems that there's no built in way to get all the pre-installed tools; the best solution seems to be to just check the related directories in the Android SDK directory.

guitcastro:
Did you try to look at se souce code of the Android Studio or the SDK Manager ? They do that.

I will try to look at it, I Just need some free time.

viktorbenei:
Based on what we found Googling around the web it seems that the SDK Manager does exactly that, it just parses the Android SDK home directories. It's mentioned that you can even install new packages just by downloading the package an placing it to the appropriate directory, and the SDK Manager will detect it correctly.

viktorbenei:
Sorry for the long delay, we just wanted to get this right.

So, it seems that the only way you can check whether a tool is installed is indeed to check the related folder. You can implement this logic, as it'll speed up the whole process/step in case the package is already installed.

That said we switched the filesystem driver from the Docker's default `Aufs` to `btrfs`, which will not fail the way `aufs` does when an android package is already installed. We're still testing the new setup, so far it seems that it'll solve the issue reliably.

We'll keep testing the system during the weekend, and if everything's ok we'll merge your step on Monday.

viktorbenei:
Bit of info about the Android SDK folder structure:

On OS X the SDK "root" is `~/~/Library/Android/sdk`, on Linux there's no canonical place, you just uncompress it somewhere. As a best practice, on bitrise.io images/VMs this location is stored in the `$ANDROID_HOME` environment. Your should add an input, with a default value `$ANDROID_HOME`, to get this location.

Once you have the root path the subfolder structure is the same on Linux & OS X:

```
$ tree -L 1 $ANDROID_HOME

/opt/android-sdk-linux
├── add-ons
├── build-tools
├── extras
├── platforms
├── platform-tools
├── SDK Readme.txt
├── system-images
├── temp
└── tools
```

If you want to check whether a build tool version is installed you can just check the directories in ` $ANDROID_HOME/build-tools`, it looks like this:

```
$ tree -L 1 $ANDROID_HOME/build-tools

/opt/android-sdk-linux/build-tools
├── 17.0.0
├── 19.1.0
├── 20.0.0
├── 21.1.2
├── 22.0.1
└── 23.0.2
```

So, if you want to check whether the `22.0.1` build-tool is installed you can just check whether the path `$ANDROID_HOME/build-tools/22.0.1` exists / is directory.

Other packages might be a bit more complicated, but the basics are the same. For example, if you want to check whether a system image is installed you can check the `$ANDROID_HOME/system-images/android-VERSION/default/IMAGE-NAME` directory.

This is how a system image directory looks like (in this case for `android-15`):

```
$ tree -L 3 $ANDROID_HOME/system-images/android-15/

/opt/android-sdk-linux/system-images/android-15/
└── default
    └── armeabi-v7a
        ├── build.prop
        ├── kernel-qemu
        ├── NOTICE.txt
        ├── ramdisk.img
        ├── source.properties
        ├── system.img
        └── userdata.img
```

`tree` can be a really useful tool to quickly scan through the structure on your own Mac/PC. With the `-L` flag you can control the maximum level of dirs it prints, and `tree` can be installed with `brew install tree` or on Ubuntu/Debian `apt-get install tree`.

For reference this is how the whole `system-images` directory is structured (these are the pre-installed version on bitrise.io at the moment):

```
tree -L 3 $ANDROID_HOME/system-images
/opt/android-sdk-linux/system-images
├── android-15
│   └── default
│       └── armeabi-v7a
├── android-17
│   └── default
│       └── armeabi-v7a
├── android-19
│   └── default
│       └── armeabi-v7a
├── android-21
│   └── default
│       └── armeabi-v7a
├── android-22
│   └── default
│       └── armeabi-v7a
└── android-23
    └── default
        └── armeabi-v7a
```

guitcastro:
That's wonderful Viktor,

I am a bit out of time until nov 30 (to finish my dissertation), but I will try to update the step ASAP with your solution ;)

viktorbenei:
Sure, no problem at all, we'll keep testing the Android system too, so far everything seems to be working so there's nothing to stop us from accepting the step in it's current state; we just have to wait for the final results, whether we can keep using the current solutions (file system).

PR #171, opened 2015-11-19
Android sdk update

guitcastro:


viktorbenei:
Can you please move your code from the `1.0.0` branch back to `master`, and just add a simple git tag with `1.0.0` as the release? This'll make it much easier for you to keep track of releases and to accept contributions / Pull Request for your step in the future.

guitcastro:
Done.

guitcastro:
Hey Viktor,

Can you check it now ? I made all the changes that you suggest.

Thanks

viktorbenei:
I'll check it ASAP, hopefully tomorrow I'll have the time.

viktorbenei:
And of course thanks for the changes & for your time! This step will be fantastic once released ;)

viktorbenei:
Note: this'll still overwrite the value provided by the user: https://github.com/guitcastro/steps-update-sdk/blob/master/step.sh#L3

viktorbenei:
FYI to test a Step easily you should run it with `bitrise` (CLI), not just the `step.sh` directly. Alternatively you can provide all the inputs as environment variables, but every variable you define in your `step.sh` will override the value defined by the user, either through `bitrise` or through an environment variable.

viktorbenei:
One more thing, you don't need to call `ls` or any loop for checking the existence of a directory (like you do here: https://github.com/guitcastro/steps-update-sdk/blob/master/step.sh#L9 ) - you can use the Bash built in operator in an `if` : `if [ -d "$DIR_PATH" ] ; then` -> the `-d` option will only be true if the path exists and it is a directory. For files you can use `-f`. So, in case of the `checkIfIsInstalled` it can be as simple as:

```
function checkIfIsInstalled {
  if [ -d "$ANDROID_HOME/$1/$2" ] ; then
    return 0
  else
    return 1
  fi
}
```

guitcastro:
@viktorbenei You are right, testing using biirise cli is more reliable and prevents me from forget to remove the variables.

A changed and tested your function (using -d instead of ls) and it's working like a charm.

A think it's finally ready to go :)

viktorbenei:
These are the things I mentioned on Slack & the ones I found:

- https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L23 -> a default value should be set, because this input has `value_options`
- https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L33 -> shouldn't the default be "none"?
- https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L49 -> `is_required` should be `false`
- https://github.com/guitcastro/steps-update-sdk/blob/master/step.yml#L59 -> `is_required` should be `false`

Finally you changed your code, so you'll have to create a new PR here - because this PR now points to an older commit hash (easiest way is to simply delete this one and just share the current version again).

viktorbenei:
Sent a PR with these fixes ;)

guitcastro:
Hey Viktor, Thanks for the PR.

My HD broken, I'm only with my notebook (which does not have linux neither os x). 
I was waiting for the repair to fix it, thanks for the PR =)

viktorbenei:
Sure, no hurry ;)

guitcastro:
I just submitted another PR. I am closing this one ;)

PR #177, opened 2015-11-28
carthage 1.0.4

vasarhelyia:
Make `platform` an input param, letting `carthage` to build for Mac as well.


viktorbenei:
Perfect, thanks for the contribution! 😉🚀

PR #220, opened 2015-12-18
flowdock-notification-step 0.1

scottrhoyt:


viktorbenei:
Thanks for the Step PR! just a couple of things: 

* you should use a full semver version number: instead of `0.1` you should use `0.1.0`
* https://github.com/scottrhoyt/bitrise-flowdock-step/blob/master/step.yml#L35 - `is_expand` should be `true`, to not to prevent using Environment Variables there (I believe the token can't/won't contain `$` anyway, so enabling `is_expand` should be safe) - you can even remove `is_expand`, true is the default
* https://github.com/scottrhoyt/bitrise-flowdock-step/blob/master/step.yml#L37 - if you set `is_dont_change_value` the user won't be able to edit it on bitrise.io / on a UI, only directly in `bitrise.yml`, so you should remove this option (default of the option is `false`)
* https://github.com/scottrhoyt/bitrise-flowdock-step/blob/master/step.yml#L23 - `is_skippable` might be better to set it to `true`. If this is `false` and this step fails it'll mark the whole build as failed; if set to `true` then only a warning will be printed in the logs.

@godrei will check the step.yml as well, but I believe these changes should be enough. BTW we always do a thorough check like this one on new steps, but accepting new versions will be much faster ;)

viktorbenei:
@godrei found one more issue: the current step ID would be `flowdock-notification-step`, which is quite verbose. It's up to you, but you can change this to be `flowdock-notification` or just `flowdock`, to make it easier to type if you're working in `bitrise.yml` directly.

viktorbenei:
These are all the issues which should be changed/reviewed (the step ID is up to you, as well as `is_skippable`, those are just suggestions), after that you'll have to share it again with the new version and (optionally) with the new step id. You can just share it the exact same way you did for this PR, if you share it with a new step ID you don't have to do any cleanup.

Note: you'll have to create a new version of the step on GitHub too, or you can delete the previous version&tag and re-create it, to point to the new `step.yml`, it's up to you.

Let us know if you need help with the changes / re-sharing!

scottrhoyt:
Thanks, @viktorbenei @godrei . I made the suggested changes and will share 0.1.1 with the step_id of `flowdock`.

viktorbenei:
awesome, we'll check it ASAP! ;)

scottrhoyt:
Closing this in lieu of #230.

PR #244, opened 2015-12-24
Xcode archive step mac

vasarhelyia:
@viktorbenei @godrei can you please take look if it seems right/if you miss something? I have some issues with that rvm workaround, but the problem might be with my setup. Still, I might need some clarification on that part to be able to test it.


vasarhelyia:
and merry Xmas 😄🎄🎁

viktorbenei:
We'll do it ASAP, thanks for the PR and Merry Christmas! 🎄😉

viktorbenei:
Sent a PR with proposed, non critical changes, but this version is perfectly acceptable as it is - merged! 🚀🎉

PR #247, opened 2016-01-01
LIFX 1.0.0

pisarm:
Hi,
 I created this simple step that communicates with a LIFX bulb using their [remote API](http://api.developer.lifx.com/). This enables the user to control the colour during builds. 

I for example, have mine set to turn purple while building, and red/green depending on the result.

Feel free to contact me with any questions.

Best regards 
 Flemming Pedersen

### DIFF
--- a/steps/LIFX/1.0.0/step.yml
+++ b/steps/LIFX/1.0.0/step.yml
@@ -1,0 +1,88 @@
+title: LIFX
+summary: Lets you get visual feedback while building, using your LIFX bulb.
+description: Longer description here
+website: https://github.com/pisarm/steps-lifx
+source_code_url: https://github.com/pisarm/steps-lifx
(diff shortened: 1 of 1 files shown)


viktorbenei:
Sounds amazing, thank you!
As this is a new step we'll do a bit more thorough review before merging your PR, but it shouldn't take more than a couple of days.

viktorbenei:
@godrei please do a review

viktorbenei:
Thank you for this awesome step! The only thing you should change is the step's ID. You should only use lowercase letters, so instead of `LIFX` you could use `lifx` as the ID. This (lowercase ID) will be a requirement in the near future (when the new `bitrise` CLI version is released).

Other than this everything's good, no other change is required.

To change the ID you just have to close this PR, and re-share the step exactly the way you did, just use lowercase ID when you specify the ID.

PR #248, opened 2016-01-02
lifx 1.0.0

pisarm:
As already described [here](https://github.com/bitrise-io/bitrise-steplib/pull/247). Id of step is now lowercase.


viktorbenei:
You might want to specify a `description` for your step (can be a copy of the `summary`) - but you can do that in an incremental step version later, we can accept this one as the first version.

Thanks for your contribution! 🙂🎉

pisarm:
Yeah - I noticed that along with another textual bug - will fix it asap.

viktorbenei:
No problem at all, sharing new versions is way quicker (as we already did our initial, more thorough review); do it when you have some time 🙂

PR #260, opened 2016-01-13
deploy-to-itunesconnect-deliver 2.2.3

blured2000:
skip_metadata and skip_screenshots are step options now.


viktorbenei:
Hi, it seems that you try to share your own fork of the Step. For ease of use we don't support this, to have multiple different sources of the same Step, unless the ownership of the Step is transferred.

If possible you should create a Pull Request to the original Step, and a new version of the official Step should be released.

PR #262, opened 2016-01-17
gradle-coveralls 1.0.0

donvigo:
This simple step sends code coverage to www.coveralls.io. Uses `./gradlew` for task execution.

### DIFF
--- a/steps/gradle-coveralls/1.0.0/step.yml
+++ b/steps/gradle-coveralls/1.0.0/step.yml
@@ -1,0 +1,37 @@
+title: Gradle Coveralls
+summary: Runs Coveralls with `./gradlew`.
+description: Sends code coverage to www.coveralls.io. Uses `./gradlew coveralls` task.
+  You have to configure [org.kt3k.gradle.plugin:coveralls-gradle-plugin](https://github.com/kt3k/coveralls-gradle-plugin)
+  in your project first.
+website: https://github.com/donvigo/steps-gradle-coveralls
+source_code_url: https://github.com/donvigo/steps-gradle-coveralls
+support_url: https://github.com/donvigo/steps-gradle-coveralls/issues
+published_at: 2016-01-17T16:15:27.886621568+02:00
+source:
+  git: https://github.com/donvigo/steps-gradle-coveralls.git
+  commit: bb3a506e707f1990c914c2a0e351ab2d251b29fa
+host_os_tags:
+- ubuntu
+project_type_tags:
+- android
+type_tags:
+- coveralls
+- gradle
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: true
+inputs:
+- gradlew_file_path: ./gradlew
+  opts:
+    description: |
+      Path for the gradlew file
+    is_expand: false
+    is_required: true
+    title: Path for the gradlew file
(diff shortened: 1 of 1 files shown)


donvigo:
I can't add labels, but please *don't merge this one*, it's blocked by this [PR](https://github.com/kt3k/coveralls-gradle-plugin/pull/50). My pull request, which add Bitrise support, was accepted, but the new version of library isn't released yet.

steps/gradle-coveralls/1.0.0/step.yml
```diff
@@ -0,0 +1,37 @@
...
+project_type_tags:
+- android
+type_tags:
+- coveralls
+- gradle
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: true
+inputs:
+- gradlew_file_path: ./gradlew
```
viktorbenei: You could set the default value to `$GRADLEW_PATH`, as you can see it in our Gradle Runner step ( https://github.com/bitrise-io/steps-gradle-runner/blob/master/step.yml#L55 ). This Env Var is configured automatically when you add a new Android project on bitrise.io, so that a user doesn't have to set the path of `gradlew` for every gradle step.

steps/gradle-coveralls/1.0.0/step.yml
```diff
@@ -0,0 +1,37 @@
...
+  opts:
+    description: |
+      Path for the gradlew file
+    is_expand: false
+    is_required: true
+    title: Path for the gradlew file
+- coveralls_task: coveralls
```
viktorbenei: Is this input required? Is there a case when the user would like to change this to something else?

steps/gradle-coveralls/1.0.0/step.yml
```diff
@@ -0,0 +1,37 @@
...
+- gradle
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: true
+inputs:
+- gradlew_file_path: ./gradlew
+  opts:
+    description: |
+      Path for the gradlew file
+    is_expand: false
```
viktorbenei: You should set `is_expand` to `true`, so that the value can be an Environment Variable (like `$GRADLEW_PATH`).

viktorbenei:
Re:Icons : if you have an SVG icon for the step you can add it yourself, or else we'll get or create one in the future.

viktorbenei:
Other then these, your step is perfect, thank you for the contribution!! 😊
If you'd decide to fix these issues for this version you can just:

* close this PR
* remove your step tag/release
* change the things
* re-create the tag/release
* re-share the step

Or, if you'd rather submit it as a new version:

* change the things
* create a new tag/release
* share the new version

donvigo:
Makes sense, I'll re-create it.

steps/gradle-coveralls/1.0.0/step.yml
donvigo: For now, only one `coveralls` Gradle task exists, but probably will appear alternative realisations. I prefer to have as many customisation points as possible :)

viktorbenei: Fair enough, I wasn't aware that it might change in the future :)

PR #297, opened 2016-02-07
applivery-deploy-ios 0.1

CesarTrigo:


steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+  git: https://github.com/applivery/steps-applivery-ios-deploy.git
+  commit: aca3a18df31de473bfc1c24b49a3d9d05a275c8e
+host_os_tags:
+- osx-10.10
+project_type_tags:
+- ios
+type_tags:
+- deploy
+- Applivery
+dependencies:
```
viktorbenei: Please use the new `deps` dependency declaration format - example: https://github.com/bitrise-io/steps-xcode-test/blob/master/step.yml#L21

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+  name: git
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: false
+run_if: ""
+inputs:
+- ipa_path: $BITRISE_IPA_PATH
+  opts:
+    description: ""
+    is_dont_change_value: false
```
viktorbenei: `is_dont_change_value: false` is the default value, you should remove it

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+    is_expand: true
```
viktorbenei: `is_expand: true` is the default value, you should remove it

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+      This is the API Key to access your account.
+
+      ## Where to get the Applivery Account API Key?
+
+      Sign in to your [Applivery.com](http://dashboard.applivery.com) account,
+      click on Developers menu option from the left side menu and copy it from the
+      Account API Key section.
+    is_dont_change_value: false
+    is_expand: true
+    is_required: false
```
viktorbenei: `is_required: false` is the default value, you should remove it

viktorbenei:
First of all thank you for the Step, we really appreciate every contribution!

Just a couple of notes, in addition to the inline comments:

* you should use a three component SemVer version instead of a two component version (`x.x.x` instead of `x.x`)
* right now the step ID is `applivery-deploy-ios` - you could drop the `deploy` part, to make it shorter/easier for those who work with the YML directly. `applivery-ios` should be enough, and you added the `deploy` `type_tag` so your Step will be included in the deploy category automatically.

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
+title: Applivery.com iOS Deploy
+summary: Deploy your awesome iOS Application to Applivery.com
+description: |-
+  Deploy an iOS application to [Applivery](http://www.applivery.com),
+    add notes and notify testers.
```
viktorbenei: Drop the additional indentation. The `description` property is handled as a Markdown text, additional indentation might change the type of the text (e.g. to `code` tag).

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+type_tags:
+- deploy
+- Applivery
+dependencies:
+- manager: brew
+  name: git
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: false
+run_if: ""
```
viktorbenei: You might want to change the `run_if` to `run_if: .IsCI` - which means that if someone runs the YML locally the step will be skipped, unless `bitrise` (CLI) run in `CI` mode.

viktorbenei: Example: https://github.com/bitrise-io/steps-hockeyapp-deploy/blob/master/step.yml#L28

steps/applivery-deploy-ios/0.1/step.yml
viktorbenei: Actually, you can drop the `git` dependency all together - `git` is a dependency of the `bitrise CLI`.

viktorbenei:
You should probably **not** print/log the `api_token`, for security: https://github.com/applivery/steps-applivery-ios-deploy/blob/master/step.sh#L31

viktorbenei:
Just a note: you might want to specify a default EnvVar for the `api_token` and `app_id` inputs. The recommendation for secrets is to store them separately from the main configuration (in Secret Env Vars on bitrise.io, and in the `.bitrise.secrets.yml` file locally). If you'd set a default Env Var for these inputs (let's say `api_token: "$APPLIVERY_API_TOKEN"`) the user can simply specify this variable somewhere else (in a secret storage). This is not a requirement, but might be a useful change - making it easier for the users to use the step "properly" / securely.

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+    title: 'Applivery: App ID'
+- notes: Deployed with Bitrise Applivery.com iOS Deploy Step.
+  opts:
+    description: Additional build/release notes
+    is_dont_change_value: false
+    is_expand: true
+    is_required: false
+    summary: ""
+    title: Notes attached to the deploy
+- notify: true
```
godrei: It's better to specify boolean values as string ("true", "false"), because some yml parser cast this bools to yes/no, which will be used as string ("yes"/"no").

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+  opts:
+    description: This flag allows you to automatically notify your testers vía email.
+    is_dont_change_value: false
+    is_expand: true
+    is_required: true
+    summary: ""
+    title: Notify Testers?
+    value_options:
```
godrei: It's better to specify boolean values as string ("true", "false"), because some yml parser cast this bools to yes/no, which will be used as string ("yes"/"no").

steps/applivery-deploy-ios/0.1/step.yml
```diff
@@ -0,0 +1,119 @@
...
+    description: Comma-separated list of tags
+    is_dont_change_value: false
+    is_expand: true
+    is_required: false
+    summary: ""
+    title: (Optional) Comma-separated list of tags
+  tags: ""
+- opts:
+    description: ""
+    is_dont_change_value: true
```
godrei: If the only one acceptable value is ios, maybe it's better to don't wire out as input, because is_dont_change_value only controls the input on workflow editor UI, but the user still able to overwrite in yml editor.

CesarTrigo:
@viktorbenei @godrei thank you so much for your comments and clarifications.
As there are some major changes, including Step ID, would it be better to Close this one and create a new Pull Request reflecting all this changes?

viktorbenei:
Correct @CesarTrigo - you should do the changes in your step's code, close this PR, and just share the new version the way you shared this one, and create a new PR.

PR #300, opened 2016-02-08
applivery-ios 0.1.1

CesarTrigo:
Improved version including Bitrise team [feedback](https://github.com/bitrise-io/bitrise-steplib/pull/297)
- Dropped dependencies
- Removed all `is_dont_change_value: false`
- Removed all `is_expand: true`
- Removed all `is_required: true`
- Shortened Step ID: `applivery-ios`
- Three component SemVer version
- Dropped the additional indentation on description
- `run_if: .IsCI`
- Not print/log API TOKEN
- Improved Secret and Env Vars
- Added support for string Booleans
- Removed `os: ios` default value and moved hardcoded to` step.sh`

### DIFF
--- a/steps/applivery-ios/0.1.1/step.yml
+++ b/steps/applivery-ios/0.1.1/step.yml
@@ -1,0 +1,93 @@
+title: Applivery.com iOS Deploy
+summary: Deploy your awesome iOS Application to Applivery.com
+description: |-
+  Deploy an iOS application to [Applivery](http://www.applivery.com),
+  add notes and notify testers.
+
+  Register a Applivery account at [http://www.applivery..com/](http://www.applivery.com)
+  and create an App to utilize this step.
+
+  You also need to get your *Account API Key* for you account and the *App Id* for the app.
+website: https://github.com/applivery/steps-applivery-ios-deploy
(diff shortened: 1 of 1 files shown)


viktorbenei:
My only note is that you should remove the `iOS` type tag (https://github.com/bitrise-io/bitrise-steplib/pull/300/files#diff-a1cd355bd80c74767c79960523ac3670R25) - it's already in the project type tags, and that should be enough (type tags are only used for categorization, project type tags are used for actually filtering steps in the list - e.g. don't show iOS steps for Android projects).

Other than that it looks perfect, and this is not a major issue, so you can fix this in the next version if you want to, you don't have to change this PR (we love incremental versions, as far as it's backward compatible ;) ).

godrei:
Looks great! :)

viktorbenei:
Merged - it's now available in the step lib, and will be available on bitrise.io in a couple of minutes, when the caches are re-generated! 🚀🎉

CesarTrigo:
Thanks million @godrei @viktorbenei!
Is there any email address to contact you in relation to the press release we'll do?

Thanks!

viktorbenei:
Thanks for the Step @CesarTrigo ! ;)
Sure -> https://www.bitrise.io/contact

PR #307, opened 2016-02-09
applivery-android 0.1.0

CesarTrigo:
Hi again guys
Now I'm performing a pull request for the Android step, forked from the iOS one. This one has been successfully tested in Bitrise.io with a Google sample App.

Please, note that I'm a little bit worried cause every time I follow the `bitrise share` steps, I found that in the process something extrange happends with the double quotes around `apk_path: $BITRISE_APK_PATH` that should look like  `apk_path: "$BITRISE_APK_PATH"`

As you can see, they are correctly written in the [original](https://github.com/applivery/steps-applivery-android-deploy/blob/master/step.yml#L27) but not [here](https://github.com/bitrise-io/bitrise-steplib/compare/master...applivery:applivery-android?expand=1#diff-c6b9d157f22170d28cc809250b822452R31)... is it normal?

Thanks and regards

### DIFF
--- a/steps/applivery-android/0.1.0/step.yml
+++ b/steps/applivery-android/0.1.0/step.yml
@@ -1,0 +1,93 @@
+title: Applivery.com Android Deploy
+summary: Deploy your awesome Android Application to Applivery.com
+description: |-
+  Deploy an Android application to [Applivery](http://www.applivery.com),
+  add notes and notify testers.
+
+  Register a Applivery account at [http://www.applivery..com/](http://www.applivery.com)
+  and create an App to utilize this step.
+
+  You also need to get your *Account API Key* for you account and the *App Id* for the app.
+website: https://github.com/applivery/steps-applivery-android-deploy
+source_code_url: https://github.com/applivery/steps-applivery-android-deploy
+support_url: https://github.com/applivery/steps-applivery-android-deploy/issues
+published_at: 2016-02-09T23:36:10.028033474+01:00
+source:
+  git: https://github.com/applivery/steps-applivery-android-deploy.git
+  commit: f76628cef4a36049026b74d0a96eabdc5d3f7e8e
+host_os_tags:
+- ubuntu
+- osx-10.10
+project_type_tags:
+- android
+type_tags:
+- deploy
+- Applivery
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: false
+run_if: .IsCI
+inputs:
+- apk_path: $BITRISE_APK_PATH
+  opts:
+    description: ""
+    is_required: true
+    summary: ""
+    title: APK file path
+- api_token: ""
+  opts:
+    description: |-
+      This is the API Key to access your account.
+
+      ## Where to get the Applivery Account API Key?
+
+      Sign in to your [Applivery.com](http://dashboard.applivery.com) account,
+      click on Developers menu option from the left side menu and copy it from the
+      Account API Key section.
+    is_required: true
+    summary: ""
+    title: Account API Key
+- app_id: ""
+  opts:
+    description: |-
+      This is the App Id that identifies your App in Applivery.com
+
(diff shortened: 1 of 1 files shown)


viktorbenei:
Hi @CesarTrigo - thanks, we'll do a review ASAP!

Re quotes and other formatting changes: don't worry. When you share your step it gets normalized, that's why quotes might get removed, empty values are replaced with `null`, comments are removed, etc. There should be no functional change, this is just the result of `stepman` reading in your YML, performing an audit on it, and then it exports in a normalized YML all the whitelisted data.

Format optimizations (like quotation removals) are done by Go's standard JSON library (official, built in JSON package - https://golang.org/pkg/encoding/json/). It should simply be a normalized version of your YML :)

CesarTrigo:
Magnificat explanation. Thank you.

steps/applivery-android/0.1.0/step.yml
```diff
@@ -0,0 +1,93 @@
+title: Applivery.com Android Deploy
+summary: Deploy your awesome Android Application to Applivery.com
+description: |-
+  Deploy an Android application to [Applivery](http://www.applivery.com),
+  add notes and notify testers.
+
+  Register a Applivery account at [http://www.applivery..com/](http://www.applivery.com)
```
viktorbenei: Typo in the `http://www.applivery..com/` text - URL is good

steps/applivery-android/0.1.0/step.yml
```diff
@@ -0,0 +1,93 @@
...
+  opts:
+    description: |-
+      This is the App Id that identifies your App in Applivery.com
+
+      ## Where to get the App Id?
+
+      Sign in to your [Applivery.com](http://dashboard.applivery.com) account,
+      click on Applications menu option from the left side menu, click on the desired App.
+      You'll find the App Id inside the (i) information block (written in red).
+    is_required: false
```
viktorbenei: `is_required: false` is the default, can be removed

viktorbenei:
I don't see any show stopper issue, @godrei will do a review too, but if you're OK with the current version we can accept it. Depends on whether you want to fix the typo in the description.

viktorbenei:
Btw the iOS version has the same link text typo https://github.com/bitrise-io/bitrise-steplib/pull/300/files#diff-a1cd355bd80c74767c79960523ac3670R7

CesarTrigo:
Thank you @viktorbenei , OK from our side and I agree that these are minor issues that can be fixed in new versions, probably next week. We are also analyzing how to do a deeper integration from Applivery to Bitrise in order to link the builds with Bitrise once the deploy is finished so I think we'll be improving both steps very often :)

godrei:
@CesarTrigo I'm ok with the PR.

Thanks for the contribution :)

viktorbenei:
You can do something similar to what Hockey does @CesarTrigo - Hockey supports "build server" and other similar properties. If I remember it correctly, if the Build Server param is set it'll be presented on the build's page on Hockey. You can get both the App's and the Build's URL with environment variables. You can find the base list of Bitrise.io Env Vars here: http://devcenter.bitrise.io/docs/available-environment-variables

viktorbenei:
🚀 Will be available on bitrise.io in a couple of minutes, after the cache refresh. Thanks for the Step! 😉

CesarTrigo:
Thanks for the clear explanation and the example from Hockey @viktorbenei we'll work on this during next week! I think everything will look very nice at the end :)
BTW, should we share our logo with you in order to be displayed in the integrations list (https://www.bitrise.io/integrations)?
Thanks!

viktorbenei:
Logo: you can add it to the step lib (to this repo) directly - just check another step how/where it have to be placed (into the step's `assets` folder) ;)

viktorbenei:
And step version updates are always welcome! ;)

PR #334, opened 2016-02-20
testfairy-deploy-android 1.0.0

rudacs:


steps/testfairy-deploy-android/1.0.0/step.yml
```diff
@@ -0,0 +1,124 @@
+title: TestFairy Deploy Android
+summary: |-
+  Deploy your awesome App to TestFairy, set Tester Groups to be notified,
+  enable email notification for the build, enable in-app notification of new
+  version, set recording duration, set if video recording is enabled and add
+  your comments and let the TestFairy do it's magic.
+description: |-
```
godrei: On bitrise.io setp's description will be displayed, maybe it would be better to provide here the same description as in summary too, to provide little bit more info about the step.
In general both bitrise-cli and bitrise.io use step's description, so its better to provide more details in description.

steps/testfairy-deploy-android/1.0.0/step.yml
```diff
@@ -0,0 +1,124 @@
...
+  To use this step you have to register
+  an account on [TestFairy.com](http://testfairy.com/).
+website: https://github.com/rudacs/bitrise-steps-testfairy-deploy-android
+source_code_url: https://github.com/rudacs/bitrise-steps-testfairy-deploy-android
+support_url: https://github.com/rudacs/bitrise-steps-testfairy-deploy-android/issues
+published_at: 2016-02-20T14:55:22.465044928-03:00
+source:
+  git: https://github.com/rudacs/bitrise-steps-testfairy-deploy-android.git
+  commit: 5e57153e55ec29675911069bbba8d439d9b5afe4
+host_os_tags:
```
godrei: Android project's workflows runs on linux vm-s, so at minimum list `ubuntu` as host_os_tags as well. xamarin projects (which can generate .apk as output) runs on os-x vm-s, so if you want to allow xamarin users to use this step then this list should contains both `osx-10.10`, and `ubuntu`.

viktorbenei:
I think it's perfect now - but please close this PR and create a new one, with only the final code in it (you can delete your branch in your fork, and just re-share the step).

PR #356, opened 2016-03-07
steps-set-xcode-git-build-number 0.1

datinc:
Add steps-set-xcode-git-build-number to StepLib


godrei:
Hi @datinc ,
thanks your PR the step looks great, however the followings should be corrected before we can merge your step.

- Step's output should be available in the next steps. Your step only prints the `BUILD_NUMBER_OUTPUT`. It should be exported with our environment manager tool (`envman`, see: https://github.com/bitrise-io/envman), it would be something like this: `envman add --key BUILD_NUMBER_OUTPUT  --value $BUILD_NUMBER_OUTPUT `

- Step's versioning follows semantic versioning, so version number should be 3 components MAJOR.MINOR.PATCH

- Step's id should not contains the `step-` prefix, in this case it should be `set-xcode-git-build-number`

Please close this PR, do the changes and share the step again by calling `bitrise run share-this-step` in your step's directory.

PR #360, opened 2016-03-08
steps-xamarin-test-cloud-for-calabash-ios 0.9.2

vasarhelyia:


viktorbenei:
wrong ID: `steps-xamarin-test-cloud-for-calabash-ios`

PR #361, opened 2016-03-08
steps-xamarin-test-cloud-for-calabash-android 0.9.2

vasarhelyia:


viktorbenei:
wrong step id `steps-xamarin-test-cloud-for-calabash-android`

PR #393, opened 2016-03-28
installrapp-deploy 1.0.0

Redth:
This adds support for InstallrApp.com as requested in https://github.com/bitrise-io/bitrise-contrib/issues/11

PR #395, opened 2016-03-29
installrapp-deploy 1.0.1

Redth:
This adds support for InstallrApp.com as requested in bitrise-io/bitrise-contrib#11

Should fix the issues from the original PR #393


viktorbenei:
Perfect, thank you!

After merge it'll be available on bitrise.io after a couple of minutes, once the caches refresh. Note: You might have to reload the workflow editor, to refresh your browser caches.

PR #396, opened 2016-03-30
set-android-manifest-package-name 0.9

kshilovskiy:
The step allows to change android manifest package name


godrei:
Hi @kshilovskiy ,
your step looks great!

The only thing you should change is the step's version, please use sematic [versioning](http://semver.org), your version should be 0.9.0.

To fix this you should just re-share your step, with the correct version:

- Close your Pull Request
- Delete your `set-android-manifest-package-name` branch in your bitrise-steplib fork
- Share the step by calling bitrise run share-this-step in your step's directory

PR #397, opened 2016-03-31
set-android-manifest-package-name 0.9.0

kshilovskiy:
The step allows to change android manifest package name


godrei:
Thanks @kshilovskiy ,
your step will be available in steplib in a hour.

PR #402, opened 2016-04-05
set-android-application-attributes 0.9.0

kshilovskiy:
This step allows to change the initial attributes of the `<application>` tag specified in AndroidManifest.xml


godrei:
Hi @kshilovskiy ,
looks greate, will be available in steplib soon!
Thanks! :)

PR #405, opened 2016-04-07
Support for fauxpasapp 0.1.1

mackoj:
This is a Step for [Faux Pas App](http://fauxpasapp.com).

### DIFF
--- a/steps/fauxpasapp/0.1.1/step.yml
+++ b/steps/fauxpasapp/0.1.1/step.yml
@@ -1,0 +1,83 @@
+title: Fauxpasapp Step
+summary: Step for Fauxpasapp
+description: |-
+  This is a Step for [Faux Pas App](http://fauxpasapp.com).
+  This is not an officially supported plugin
+website: https://github.com/mackoj/fauxpasapp-bitrise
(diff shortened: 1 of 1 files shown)


viktorbenei:
Sorry, we somehow missed a couple of things:

`fauxpas_project_path` & `fauxpas_workspace` - you should use a single input -> the `$BITRISE_PROJECT_PATH` environment can be either a project or a workspace file, it's not stored in separate Env Vars. The name is a bit confusing, sorry, but we didn't want to introduce two separate Env Vars, to reduce the number of params overall.

A note about `is_always_run` : you should probably not set this to true, as that'll mean that the step will run even if Git Clone fails. This flag is designed for notification steps mostly, so that e.g. a Slack message will be sent even if everything else failed. For the same reason you should also remove `run_if: "true"`, you should use the default error handling provided by the CLI instead.

You should probably remove the `fauxpas_source_dir` input too, as the step will run in `$BITRISE_SOURCE_DIR` by default - but this is up to you, if you want to keep this input you're free to do so.

I swear these are the last things, everything else looks perfect! ;)

PR #430, opened 2016-04-21
Add new step: xcode-project-info 1.0.0

tadija:
Hello Bitrise,

I've made this step according to our talk in https://bitrise.uservoice.com/forums/235233-general/suggestions/11312937-version-and-build-number-as-environment-variables

I hope everything's fine, it seems to work in my local and 'remote' tests.

Cheers,
// T

### DIFF
--- a/steps/xcode-project-info/1.0.0/step.yml
+++ b/steps/xcode-project-info/1.0.0/step.yml
@@ -1,0 +1,36 @@
+title: Xcode Project Info
+summary: Extracts version and build number to environment variables.
+description: |-
+  This step will simply read version and build number from given `Info.plist` path,
+  then export those to **APP_VERSION** and **APP_BUILD** environment variables.
+
+  After this you can use these variables in other steps (ex. sending message on Slack).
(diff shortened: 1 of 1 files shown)


viktorbenei:
Awesome!

Just a couple of notes:

* You should declare the `outputs` the Step generates
* `type_tags` should be `xcode` or something similar instead - you can check what other steps use at http://www.bitrise.io/integrations
* please use the new `deps` dependency declaration format, instead of the (now deprecated) `dependencies` format
* please remove `run_if` if you don't want to declare a run if expression
* remove `is_dont_change_value` - it should be `false` not `true`, and `false` is the default value of this option
* you should also remove `is_expand: true`, as that's the default value
* as well as `value_options: []`, if you don't want to provide value options (I guess you don't ;)

viktorbenei:
The fastest way to fix these: just do the fixes, bump your version number and share it as `1.0.1`. We'll accept that one (if everything's OK) and just close & not accept this one

tadija:
Thanks, I just did that!

PR #431, opened 2016-04-21
Add new step: Xcode project info 1.0.1

tadija:
Hello, here's a new version with minor fixes suggested in #430.

PR #432, opened 2016-04-21
Add new step: xcode-project-info 1.0.1

tadija:
Ok, here's another take on #430 and #431. :)

### DIFF
--- a/steps/xcode-project-info/1.0.1/step.yml
+++ b/steps/xcode-project-info/1.0.1/step.yml
@@ -1,0 +1,40 @@
+title: Xcode Project Info
+summary: Extracts version and build number to environment variables.
+description: |-
+  This step will simply read version and build number from given `Info.plist` path,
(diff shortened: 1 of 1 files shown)


viktorbenei:
Perfect!

Last question: are you sure that you want to use this ID for your step? Once it's shared there's no option to change it, only to share it under a new name and deprecate the old one - to not to break user workflows who already use the step.

tadija:
Thanks for notice, but that was mostly the reason I named in a little bit broader sense - as sometimes later, it can be extended for reading project info other then just build or version number.

viktorbenei:
Fair enough, and thank you for the contribution! Your step is now available in the Step Lib - it'll appear on bitrise.io too in 5-10 minutes, once the caches pick up the change ;)

tadija:
Awesome!

Thanks for the very quick feedback on this.

viktorbenei:
Thanks for the Step! 😉

PR #435, opened 2016-04-25
post-jira-comment-with-build-details 0.1.1

dag-io:


godrei:
Looks great to me! 
What do you think @viktorbenei?
Thanks @dag-io !

viktorbenei:
@dag-io are you sure that you don't want this step to run for Pull Request builds (https://github.com/bitrise-io/bitrise-steplib/pull/435/files#diff-687020de9cbe47e8d7720f9a1c04a541R18) ?

dag-io:
@viktorbenei Yes because I'm afraid it would generate "duplicates" builds. I thought the build for the **push** would be the one published in JIRA. What do you think ?

viktorbenei:
Can be, it's not an issue at all, just want to make sure that this is what you intended to do ;)

viktorbenei:
you might also want to set more specific `type_tags`, but that's also up to you

viktorbenei:
if you say it's good as it is, we can merge it

viktorbenei:
you can of course fix/change this after sharing too, by sharing a new version ;)

viktorbenei:
merged - thanks for the contribution! We'd suggest you to use a more specific `type_tags` in your next version, something like `jira`, `issue-tracker`, to make it easier to find your step in the lib

dag-io:
Changed applied here : bitrise-io/bitrise-steplib#437

viktorbenei:
Awesome, thank you! (just merged it 😉 )

PR #440, opened 2016-04-27
post-jira-comment-with-build-details 0.1.4

dag-io:
- Rewritten in PHP. Bash part similar from https://github.com/bitrise-io/steps-slack-message)
- Add dependencies for PHP both with homebrew and apt-get (but did not check with Android project. I will fo it later)
- Add output variable (JIRA_ISSUE_KEY)
- Rewrite "run_if" condition
- Add tags

### DIFF
--- a/steps/post-jira-comment-with-build-details/0.1.4/step.yml
+++ b/steps/post-jira-comment-with-build-details/0.1.4/step.yml
@@ -1,0 +1,86 @@
+title: Create JIRA comment with build
+summary: Will post a comment on the issue page with build details (download link,
+  build number, etc)
+description: Will post a comment on the issue page with build details (download link,
+  build number, etc)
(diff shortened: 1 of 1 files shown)


viktorbenei:
Awesome!

Just one note for this change:

```
< - jira_domain: null
---
> - jira_url: null
```

You should try to avoid changing the ID of the inputs, as changing it can break existing workflows where the user specified the value for the previous input ID (`jira_domain` in this case), and after the step version upgrade the step won't work as the input now expects a new input ID (`jira_url`).

As your step is still in beta (0.x) it's acceptable I think, just wanted to note this.

If you say it's OK this way, we can merge the step, it looks good otherwise ;)

dag-io:
Yes I did it on purpose. I use sementic versioning. From 1.0 there will be no BC break ;-)

viktorbenei:
Fair enough ;)

PR #444, opened 2016-04-30
Step - Set iOS Bundle Identifier v1.0.1

jsinh:
Step - Set iOS Bundle Identifier v1.0.1
Please review and accept the PR, thanks!


viktorbenei:
Thanks for the contribution! 😉

PR #445, opened 2016-04-30
set-ios-bundle-identifier 1.0.2

jsinh:
Fixed info.plist path in quotes to allow spaced file path values, thanks!


viktorbenei:
Awesome, thank for the update/fix!

PR #447, opened 2016-05-01
Fix icon alignment

dag-io:
Now it as an 1:1 aspect to I guess it will display correctly on bitrise.io


viktorbenei:
Thanks for the fix ;)

PR #463, opened 2016-05-11
profile-downloader 1.0.0

mAu888:
# What's in here

This adds the possibility to download a current provisioning profile using [sigh](https://github.com/fastlane/fastlane/tree/master/sigh). It exports all downloaded variables to `$provisioning_profile_paths`, which can then be used in the `Certificate & profile installer` step.

Ad hoc profiles will always be regenerated, in order to include all available devices.

### DIFF
--- a/steps/profile-downloader/1.0.0/step.yml
+++ b/steps/profile-downloader/1.0.0/step.yml
@@ -1,0 +1,53 @@
+title: Profile downloader
+summary: Provisioning profile downloader for iOS projects
+description: Use this step to automatically download provisioning profiles for your
+  project. This step uses [sigh](https://github.com/fastlane/fastlane/tree/master/sigh)
+  from the fastlane toolchain.
+website: https://github.com/mAu888/steps-sigh
+source_code_url: https://github.com/mAu888/steps-sigh
+support_url: https://github.com/mAu888/steps-sigh/issues
+published_at: 2016-05-11T16:44:16.222754322+02:00
+source:
+  git: https://github.com/mAu888/steps-sigh.git
+  commit: bada07f5534e9a7915e0ed5e89eaa77ca6e2d568
+host_os_tags:
+- osx-10.10
+project_type_tags:
+- ios
+type_tags:
+- provisioning
+- code signing
+- fastlane
+is_always_run: false
+is_skippable: false
+run_if: ""
+inputs:
+- opts:
+    is_required: true
+    title: Provisioning profile type
+    value_options:
+    - adhoc
+    - appstore
+    - development
+  profile_type: adhoc
+- bundle_identifier: null
+  opts:
+    is_required: true
+    title: Application bundle identifier
+- certificate_id: null
+  opts:
+    is_required: true
+    summary: The identifier of the certificate associated with the provisioning profile
+    title: Certificate ID
+- opts:
+    is_required: false
+    summary: Team ID, if you are in multiple development teams.
+    title: Team ID
+  team_id: null
+outputs:
+- opts:
+    is_dont_change_value: true
+    summary: The provisioning profiles that were found and downloaded. The variable
+      may be used in the `Certificate and profile installer` step.
+    title: Available provisioning profiles
+  provisioning_profile_paths: null


viktorbenei:
Wow, it sounds super awesome! We'll review it ASAP!

godrei:
Hi @mAu888 ,
can you please change the output env to upper case (`provisioning_profile_paths` -> `PROVISIONING_PROFILE_PATHS`), it would better fit to steplib.

If step's main goal is to use together with `certificate-and-profile-installer` step, then your step's output could be named like `certificate-and-profile-installer` step's input (`- provisioning_profile_url: $BITRISE_PROVISION_URL`), namely: `BITRISE_PROVISION_URL `.  
This way you do not need to modify the `certificate-and-profile-installer` step's `provisioning_profile_url` input's value.

Another thing, if i'm rigth your step downloads the profiles and export the [paths](https://github.com/mAu888/steps-sigh/blob/master/collect_profiles.swift#L19-L21). `certificate-and-profile-installer` step requires `file://` suffix in local paths, so i think you should add `file://` prefix to local file paths.

mAu888:
Hi @godrei,
Regarding the rename of the output to `BITRISE_PROVISION_URL`: doesn't stepman overwrite the value in `BITRISE_PROVISION_URL`, if I assign my content? It would probably be advisable to concaternate, if that variable exists (e.g. `if [ ! -z "${EXPORT_PROFILES_PATHS}" ]; then EXPORT_PROFILES_PATHS="${EXPORT_PROFILES_PATHS}|${downloaded_paths}" fi`), or?

The step already returns `file://` prepended paths. Somehow I think the steps in the steplib should be unified on how they handle paths in general. While debugging my step I tried to deploy the downloaded profile using `Deploy to bitrise` step, but that one is for example not able to handle `file://` paths. Maybe there's also a good reason why some steps handle paths differently?

viktorbenei:
"Maybe there's also a good reason why some steps handle paths differently?" - in general, when an input can be both a remote URL or a local file path we now use the `file://` prefix for local paths. If the step only supports local paths the `file://` scheme is not required, as there's nothing to decide (about how to get the file).

But I agree, we could probably do it better - and we will, eventually ;)

godrei:
@mAu888 no, it does not overwrite. Actually [envman](https://github.com/bitrise-io/envman) handles bitrise env vars. If you upload prov profile then it will be downloaded and it's path will be exported into `BITRISE_PROVISION_URL` in the begining of the build, always the last defined `BITRISE_PROVISION_URL` value will be used, so you can overwrite it. 

Regarding to concatenate, i do not see, why is it a use case. I mean, if i upload my profiles, then i do not need the step, or if i use the step, then i will not upload profiles.

mAu888:
Depends, I could imagine for example having multiple profiles (for today extensions, apple watch etc.) which currently would require this step multiple times.

godrei:
I do not think it's a good idea to upload a part of the required profiles to bitrise and the other part not.

By the way, your step downloads 1 profile only? So you have to use more times the step, to download all of the required profiles?

If so, then concatenate is required to easy achieve that.

mAu888:
Well it downloads all profiles for a given bundle identifier. But for extensions you will have a different bundle identifier. Could be a future improvement to provide a list of bundle identifiers to the step instead.

godrei:
Ok, so i agree there is a use case to concatenate `BITRISE_PROVISION_URL`s.

viktorbenei:
Hmm.. I think we forgot about this PR.. Checking the things again it seems OK for me.

@godrei @mAu888 should we merge?

viktorbenei:
Hi @mAu888 ,

The step & PR seems to be good, but in the meantime we updated our guidelines. Can you please create a new PR, or accept the points here? Thank you!

https://github.com/bitrise-io/bitrise-steplib/blob/master/PULL_REQUEST_TEMPLATE.md

- [ ] I read the [Step Development Guideline](https://github.com/bitrise-io/bitrise/blob/master/_docs/step-development-guideline.md)
- [ ] I did run `stepman audit --step-yml ./step.yml` (in the step's repository - note, `stepman` is part of the Bitrise CLI, you don't have to install it separately)
- [ ] I read and accept the [Abandoned Step policy](https://github.com/bitrise-io/bitrise-steplib#abandoned-step-policy)

mAu888:
I'm currently short on time but I'll reopen the PR as soon as I found time to revisit the required steps! Should be latest after my vacation 🌴 , so in about three weeks.

viktorbenei:
Awesome, let us know if we can help with anything!

PR #464, opened 2016-05-12
web-translate-it 1.0.2

mAu888:
This fixes an issue that the current directory was not taken as a working directory, if no working directory was given. My bad 🙈.


viktorbenei:
Thanks for the fix and sorry for the long delay, the PR somehow slipped through our notifications.. :/

PR #506, opened 2016-06-08
Fix append version config bug

BOOMik:
Oops. Fix bug in config


viktorbenei:
No problem, thanks for the quick fix! 😉

PR #516, opened 2016-06-14
change-value 1.0.0

BOOMik:


steps/change-value/1.0.0/step.yml
```diff
@@ -0,0 +1,56 @@
...
+  git: https://github.com/BOOMik/step_change_value.git
+  commit: 8a9481ebe474510243e9fe7fda62d2dcba63731e
+type_tags:
+- build
+- utils
+- xamarin
+- ios
+- android
+- code
+run_if: .IsCI
```
viktorbenei: why `.IsCI`?

viktorbenei:
LGTM, the only thing what looks odd is `run_if: .IsCI` - I think this should be left on the default value instead, as the step can be useful in non CI mode too, and the user can change `run_if` if wants to

steps/change-value/1.0.0/step.yml
```diff
@@ -0,0 +1,56 @@
+title: Change value in file
+summary: Change value in file. (e.g. constants)
+description: Change value in file. (e.g. constants)
+website: https://github.com/BOOMik/steps-set-ios-version
```
godrei: it looks website url is not correct. should be: 'https://github.com/BOOMik/step_change_value` instead of: `https://github.com/BOOMik/steps-set-ios-version`.

also ensure: `source_code_url ` and `support_url `

Am i missing something?

BOOMik:
Thx for notes. I fix it. (Or need new PR?)

godrei:
@BOOMik you need to do a new PR. so close this PR, delete the branch and then re-share it. thanks

BOOMik:
Ok

PR #544, opened 2016-07-02
Pebble build

HBehrens:


HBehrens:
use https://github.com/HBehrens/bitrise-step-pebble-playground to try this step

viktorbenei:
Awesome, we'll do a review ASAP! (first review takes a bit longer, future step updates are usually just a couple of hours)
Thanks for the new step!! ;)

steps/pebble-build/1.0.0/step.yml
```diff
@@ -0,0 +1,75 @@
...
+      https://developer.getpebble.com/legal/terms-of-use
+
+      PEBBLE DEVELOPER LICENSE
+      https://developer.getpebble.com/legal/sdk-license
+    is_expand: true
+    title: Accept Pebble TERMS OF USE and PEBBLE DEVELOPER LICENSE?
+    value_options:
+    - "yes"
+    - "no"
+  sdk_accept_legalese: "no"
```
godrei: are you sure that, here "no" is the correct default value? i think you can not use the tool without agree the terms.

steps/pebble-build/1.0.0/step.yml
```diff
@@ -0,0 +1,75 @@
...
+- opts:
+    description: |-
+      To use the Pebble SDK, you must agree to the following:
+
+      PEBBLE TERMS OF USE
+      https://developer.getpebble.com/legal/terms-of-use
+
+      PEBBLE DEVELOPER LICENSE
+      https://developer.getpebble.com/legal/sdk-license
+    is_expand: true
```
godrei: you can remove this line, `is_expand` is true by default (no re-share required, just for later releases)

steps/pebble-build/1.0.0/step.yml
HBehrens: Automatically answering this question with "YES" defeats the purpose of this legal question. Originally, I wanted to leave the answer open, but when publishing the linter told me that I have to pick a value from `value_options`. "NO" was the only viable answer to me then.
I hope that the error message you get when you use this step out of the box explains the required actions well enough.

steps/pebble-build/1.0.0/step.yml
HBehrens: I was hoping that `is_expand` would keep this section open when you look at the workflow composer. Unfortunately, I wasn't able to test this during development as Bitrise doesn't show the visual composer until a step is published.
The purpose of expanding this section is to make the usage of the step (and the required action to pick "YES") as frictionless as possible.

steps/pebble-build/1.0.0/step.yml
godrei: What do you think about this: 
- remove the value options
- leave the default value empty
- mark the input as required (`is_required: true`)
- complete the description with something like: __You have to provide `yes` as input value to accpect the licence.__

The only problem with default `false` is that users will not read the whole step description, and probably will skipp to set this input and build will fail (hopefully just once).

If you follow my mentions the bitrise UI will mark the missing input and probably this way more users will provide the desired input at first.

steps/pebble-build/1.0.0/step.yml
godrei: `is_expand` true means that this input's value should be used as environment variable, for example:
if $HOME env is available in the system, if you pass $HOME as input value (`sdk_accept_legalese: $HOME`), then the step will become the value of HOME environment.

HBehrens: I misunderstood the purpose of this value then. Is there any page that explains all the possible options? I basically looked at the current library of steps and made a few guesses.

steps/pebble-build/1.0.0/step.yml
HBehrens: Makes sense, will change.

godrei:
Hi @HBehrens ,
the step looks great now!

You need to re-share the step, because `source: commit:` is set during the share process, currently it points to the previous state.

How to re-share:
- close this PR
- delete `pebble-build` branch from your steplib fork
- share again the step

HBehrens:
As I am a believer in preserving history of both commits and discussions, I tried to adjust the commit hash and date manually but this caused the build checks to fail. Unfortunately, I don't have the permission to read the actual output of those failing tests.

Is there a reasons why I need to republish the step instead of adjusting date/hash manually?

viktorbenei:
> Unfortunately, I don't have the permission to read the actual output of those failing tests.

The issue is the mismatch between the commit hash and the tag in your repository. I'd advise you to just create a new version (tag) of your step and then share that one, there's no requirement that the first version you share should be 1.0.0, in fact we usually share it as 0.0.1 or 0.9.0 even our own steps, but it can be any version - in your case most likely 1.0.1

viktorbenei:
> As I am a believer in preserving history of both commits and discussions

Feel free to link from the other PR to this discussion, GitHub is quite clever about figuring out the links between PRs :)

HBehrens:
Sorry, it was way easier to just edit the commit hash in the JSON.

viktorbenei:
No problem, everything seems to be OK now, thank you for the contribution!

PR #579, opened 2016-07-24
Nvm

Almouro:
Hi guys!

I've made the step work on Ubuntu and improved my test.
Plus I added an icon :)


steps/nvm/1.0.0/step.yml
```diff
@@ -0,0 +1,46 @@
...
+support_url: https://github.com/almouro/bitrise-nvm-step/issues
+published_at: 2016-07-24T19:29:12.711068565+02:00
+source:
+  git: https://github.com/Almouro/bitrise-nvm-step.git
+  commit: 215a76f4a17401e5cb42d079c7321e37780f7c47
+host_os_tags:
+- osx
+- ubuntu
+project_type_tags:
+- react-native
```
viktorbenei: we don't have react-native as a project type yet - this means that this step won't show up for any project by default, until you turn off the "filter by project type" toggle in the step list. This step I think should be general enough, so you should simply remove the `project_type_tags` list - that means that the step can work with any project type. Of course, if that's not the case, you can keep it.

viktorbenei:
The only minor issue was `project_type_tags`, which is not a show stopper - if you'd want to "fix" it, feel free to do that and release an update.

Thanks for the contribution! 🚀

PR #648, opened 2016-09-16
Amazon s3 uploader

scruffyfox:

Now in its own repository ;)

### DIFF
--- a/steps/amazon-s3-uploader/1.0/step.yml
+++ b/steps/amazon-s3-uploader/1.0/step.yml
@@ -8,7 +8,7 @@ description: |-
 website: https://github.com/scruffyfox/steps-amazon-s3-uploader
 source_code_url: https://github.com/scruffyfox/steps-amazon-s3-uploader
 support_url: https://github.com/scruffyfox/steps-amazon-s3-uploader/issues
-published_at: 2016-09-16T14:54:39.39399456+01:00
+published_at: 2016-09-16T14:56:13.213374723+01:00
 source:
   git: https://github.com/scruffyfox/steps-amazon-s3-uploader.git
   commit: 4047a517e1af1f50e69a41873c4844cfe3bb5449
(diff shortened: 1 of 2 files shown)


steps/amazon-s3-uploader/1.0/step.yml
```diff
@@ -0,0 +1,105 @@
...
+source:
+  git: https://github.com/scruffyfox/bitrise-amazon-s3-uploader.git
+  commit: 1538d8a978516687b098c52e2a7812ca9424cf68
+type_tags:
+- deploy
+- amazon
+- s3
+- upload
+- uploader
+deps:
```
viktorbenei: please replace `deps` to work on both Mac OS and Linux with:

```
deps:
  brew:
  - name: awscli
    bin_name: aws
  apt_get:
  - name: awscli
    bin_name: aws
```

steps/amazon-s3-uploader/1.0/step.yml
```diff
@@ -0,0 +1,105 @@
...
+  brew:
+  - name: awscli
+is_requires_admin_user: true
+is_always_run: false
+is_skippable: true
```
viktorbenei: why is this skippable? if skippable is true then even if the step fails the build will continue

viktorbenei:
Sorry, we just found these issues - if you want to we can accept this version, and then you can publish an update, or you can fix it and share it again.

Which one would you prefer?

scruffyfox:
That's ok i'll fix and update here, im still new to the steps spec so didnt know what values to put :)

steps/amazon-s3-uploader/1.0/step.yml
scruffyfox: Yes, i left it like this because in our use case, its not always required for the upload to happen, if it doesnt it shouldn't fail the build, but I can change this if it makes more sense?

viktorbenei:
Actually I'd advise you to do this:

1. fix it in your step
2. create a new version there, don't delete the current one (easier and no need)
3. then share the new version

steps/amazon-s3-uploader/1.0/step.yml
viktorbenei: I think the default should still be `false`, and most users / use cases I think would expect the step to cause the build to fail if the step fails

You can always just click the toggle on Bitrise.io or set `is_skippable` in your `bitrise.yml`

scruffyfox:
Sorry, I tried to make a new pull request but github updated this one

viktorbenei:
Right, I forgot that, sorry - just delete your current branch in your steplib fork, and run share again :)

viktorbenei:
By current branch I mean the `amazon-s3-uploader` branch in your `bitrise-steplib` fork repo

PR #666, opened 2016-09-23
aws-device-farm-runner 0.0.2

fadookie:
Third time's the charm?


viktorbenei:
Awesome, thank you so much for sharing this step!!

Sorry for the long acceptance process, but I promise that sharing new versions will be way faster in the future (just don't delete or change release/version tags ;) )!

This is mainly because when you share an update for a step our system prints us the summary of the diff compared to the previous version, so we only have to do the full thorough "audit" the first time, after that it's just checking the diffs ;)

Thanks again for your contribution, you rock! 🚀

PR #700, opened 2016-10-19
android-info 1.0.1

radimvaculik:

### DIFF
--- a/steps/android-info/1.0.1/step.yml
+++ b/steps/android-info/1.0.1/step.yml
@@ -1,0 +1,51 @@
+title: Android info
+summary: Android APK info provider
+description: Provides all possible Android APK information as package name, version
+  name or version code.
+website: https://github.com/thefuntasty/bitrise-step-android-info
+source_code_url: https://github.com/thefuntasty/bitrise-step-android-info
+support_url: https://github.com/thefuntasty/bitrise-step-android-info/issues
+published_at: 2016-10-19T14:14:42.059481445+02:00
+source:
+  git: https://github.com/thefuntasty/bitrise-step-android-info.git
+  commit: 9d343a42df2b8a3b64ec328b236f1fc609aaf0d8
+project_type_tags:
+- android
+type_tags:
+- artifact
+is_requires_admin_user: false
+is_always_run: true
+is_skippable: false
+run_if: .IsCI
+inputs:
+- apk_path: $BITRISE_APK_PATH
+  opts:
+    description: |
+      File path to APK file to get info from.
+    is_required: true
+    title: APK file path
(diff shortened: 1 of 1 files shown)


steps/android-info/1.0.1/step.yml
```diff
@@ -0,0 +1,51 @@
...
+  git: https://github.com/thefuntasty/bitrise-step-android-info.git
+  commit: 9d343a42df2b8a3b64ec328b236f1fc609aaf0d8
+project_type_tags:
+- android
+type_tags:
+- artifact
+is_requires_admin_user: false
+is_always_run: true
+is_skippable: false
+run_if: .IsCI
```
viktorbenei: I'd remove this, this step seems to be useful in non CI environment too

steps/android-info/1.0.1/step.yml
```diff
@@ -0,0 +1,51 @@
...
+published_at: 2016-10-19T14:14:42.059481445+02:00
+source:
+  git: https://github.com/thefuntasty/bitrise-step-android-info.git
+  commit: 9d343a42df2b8a3b64ec328b236f1fc609aaf0d8
+project_type_tags:
+- android
+type_tags:
+- artifact
+is_requires_admin_user: false
+is_always_run: true
```
viktorbenei: I'm not sure about this - if a previous step failed I'd say there is no advantage of running this step

steps/android-info/1.0.1/step.yml
```diff
@@ -0,0 +1,51 @@
+title: Android info
```
godrei: `title` should be `APK info` instead of `Android info`, because `APK info` would better explain, that this step will provides infos about apk file/files.

steps/android-info/1.0.1/step.yml
radimvaculik: Yes, make sense. Thanks.

steps/android-info/1.0.1/step.yml
radimvaculik: Yep, you're right.

steps/android-info/1.0.1/step.yml
godrei: i think `is_always_run` should be `false`, because `true` would mean, that the step should run even the apk generation would fail.

steps/android-info/1.0.1/step.yml
godrei: `run_if: .IsCI` means the step should run __only__ if `bitrise-cli` runs in CI mode (for example on website), so locally the step will be skipped (if CI mode not set by hand).

PR #714, opened 2016-10-20
Yarn 0.0.3

jonoirwinrsa:


steps/yarn/0.0.3/step.yml
```diff
@@ -0,0 +1,45 @@
...
+type_tags:
+- react-native
+- npm
+- yarn
+deps:
+  brew:
+  - name: node
+  - name: yarn
+is_always_run: false
+is_skippable: true
```
viktorbenei: are you sure about this? is skippable: true means that if the step fails that should not mark the build as failed

this is useful for e.g. IM (Slack) steps where you don't want to mark the build as failed if the step fails, but I guess this should not be used for this step

viktorbenei: Nah, the user can overwrite any property of the step! ;)

The things you specify in your `step.yml` are the defaults, but the user can overwrite any of the props in the `bitrise.yml`

PR #925, opened 2017-03-22
Cordova ios 0.2.0

nhammond101:

### DIFF
--- a/steps/cordova-ios/0.2.0/step.yml
+++ b/steps/cordova-ios/0.2.0/step.yml
@@ -1,0 +1,59 @@
+title: Cordova iOS
+summary: This step allows you to build cordova-based iOS projects.
+description: |-
+  The Cordova CLI is the main tool to use for the cross-platform workflow
+  described in the [Overview](https://cordova.apache.org/docs/en/latest/guide/overview/index.html).
+website: https://github.com/waracle/steps-cordova-ios
+source_code_url: https://github.com/waracle/steps-cordova-ios
+support_url: https://github.com/waracle/steps-cordova-ios/issues
+published_at: 2017-03-22T14:32:20.559909976Z
+source:
+  git: https://github.com/Waracle/steps-cordova-ios.git
+  commit: 85ebaa7e3859f1800d79d9df7d77a2ad1c21a1ae
+host_os_tags:
+- macosx
+type_tags:
+- cordova
+deps:
+  brew:
+  - name: npm
+inputs:
+- cordova_dir: $CORDOVA_APP_PATH
+  opts:
+    description: Change the base directory to run Cordova from.
+    is_required: false
+    title: The directory of the Cordova application.
+- cordova_command: build
+  opts:
+    description: Run `cordova --help` to check list of available commands.
+    is_required: true
+    title: Cordova command to run.
+    value_options:
+    - build
+    - platform add
+- opts:
+    description: Choose which platform to to run the command for.
+    is_required: true
+    title: Platform to build.
+    value_options:
+    - android
+    - ios
+  platform_name: ios
+- build_options: --verbose
+  opts:
+    description: |-
+      Options added to the end of the Cordova call.
+      You can use multiple options, separated by a space
+      character. For example: `--verbose`; `--debug`
+      `--release --buildConfig=..\myBuildConfig.json`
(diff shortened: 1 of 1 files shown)


godrei:
Hi @nhammond101 ,

thanks for the new step!

Currently we are working on cordova and react-native support on bitrise. Because of cordova support we have a cordova step in progress. I would be really happy if we could work together. We created a github [organisation](https://github.com/bitrise-community), where we will store the steps, on which we are working with the community together, read more on [discuss](https://discuss.bitrise.io/t/bitrise-steps-community/496/7).

So what do you think about to move your step repository to the `bitrise-community` organisation, we will invite you as a collaborator if you wish to contribute in the future. This way i will be able to extend your step with our requirements as well.

Also i'm not sure why you limited this step to ios? The step title is `Cordova iOS`, but the step has an input `platform_name`. I think we should create a platform unified cordova step, instead of creating one step per platform.

What do you think?

nhammond101:
Hi @godrei,

Thanks for the consideration, I'd be more than happy to transfer the repo and would be eager to contribute going forward.

As for limiting this step to iOS, that was only to do with the underlying host OS, as there's a Cordova Android (https://github.com/vgaidarji/steps-cordova-android) step that wasn't compatible with building on OSX.  This PR added Android as a Cordova platform, and I only limited adding Android as I didn't have access to properly test the other Cordova platforms fully. 

I'd be really interested in trying to unify/create a single step that would build on both stacks

godrei:
Hi @nhammond101 ,
sorry for my late response!

During a new platform support, we first generate a sample project, then we create a how-to guide which uses our existing steps if possible, then we develop the required steps and add the new platform specific scanner to our init tool (which generates the default configuration at Add New App page on bitrise.io).

In Cordova support my progress is the following:

- sample app: https://github.com/bitrise-samples/sample-apps-cordova
- how-to guide: https://discuss.bitrise.io/t/how-to-build-cordova-project/1312

Currenty i solved the build with our `Xcode Archive for iOS` and `Gradle Runner` steps. I'm just wondering if we should build the cordova projects with the preiously mentioned steps, or it would be better to use a cordova builder step (which uses the cordova cli).

Advantages of using the existing steps:

- there are a lots of edge case handled in the existing steps (like exporting the outputs, filtering the outputs, exporting the desired ipa from the xcarchive, etc...)

Currently my idea is the following:

- the scanner will check if the generated projects (located under `/platforms` directory) are under source controll, if they are gitignored we add an extra step which runs `cordova prepare` to generate the projects (in case of android project we also need to generate the gradle wrapper, since it is not generated by `cordova prepare`)
- once the projects are available on the build vm we can use our existing ios and android scanner to collect the required inputs for `Xcode Archive for iOS` and `Gradle Runner` steps.

Are you agree that it is a good idea to use the mentioned steps, or do you know any reason, which would require to build with the cordova-cli?

As far as i know there is no default testing library in cordova projects (i seen that the cordova team is using Jasmin), what do you think is Jasmin commonly used to test Cordova projects?

nhammond101:
Hi @godrei 

Sorry for my slow response, managed to find some time for a holiday!  

I agree with you that reusing the existing steps is the way forward, given that they were developed from real usage and the Cordova CLI basically calls them under the hood anyway.  

At Waracle, we've been using Karma/Jasmine in our Cordova projects, so it seems a sensible place to start with Jasmine for testing.

What are the next steps, I'm eager to help in any way I can.

godrei:
Hi @nhammond101 ,
sorry again for my late response.

We decided to use the cordova-cli to build cordova projects, instead of solving the build with our existing step, since we faced with a lots of edge cases.

I would have some questions before we release our cordova support. I created a topic on our discuss page: https://discuss.bitrise.io/t/state-of-the-cordova-scanner/1555

I would really appreciate, if you could take a look on the topic and let me know what do you think about the scanner.

viktorbenei:
@godrei should we close this PR? 

Please also make sure that our unified step is in the bitrise community github organization! 

@nhammond101 let me know if you'd like to join https://github.com/bitrise-community and I'll invite you 😉

nhammond101:
@godrei i've replied on the cordova scanner discuss page, tl;dr looks good, just had a question about the apk keystore

@viktorbenei that would be perfect, thanks :)

PR #1023, opened 2017-05-26
codified_security_bitrise step icon added

adamivancza:
Like this? :)


viktorbenei:
right now we only accept SVG icon images, sorry I missed this the first time

adamivancza:
Ahh okay - I'll try to get an SVG

viktorbenei:
Thank you @adamivancza ! :)

PR #1045, opened 2017-06-07
hockey-multi-deploy 1.0.1

johsoe:

### DIFF
--- a/steps/hockey-multi-deploy/1.0.1/step.yml
+++ b/steps/hockey-multi-deploy/1.0.1/step.yml
@@ -1,0 +1,58 @@
+title: Hockey CI Multi Deploy
+summary: Step for deploying one or more builds to hockey
+description: "Deploys any build artifacts and uploads to Hockey. \nAccepts a json
+  structure with hockey id, app id and a file path to the build.\nRead more on the
+  github page."
+website: https://github.com/nodes-android/ci-bitrise-deploy-step
+source_code_url: https://github.com/nodes-android/ci-bitrise-deploy-step
+support_url: https://github.com/nodes-android/ci-bitrise-deploy-step/issues
+published_at: 2017-06-07T17:54:32.999157409+02:00
+source:
+  git: https://github.com/nodes-android/ci-bitrise-deploy-step.git
+  commit: 8451bf58a361b11f5d4254c9a0d7aec74ea20d0a
+host_os_tags:
+- osx-10.10
+project_type_tags:
+- android
+- ios
+type_tags:
+- script
+deps:
+  brew:
+  - name: git
+  - name: wget
+  - name: ruby
+  apt_get:
+  - name: git
+  - name: wget
+  - name: ruby
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: false
+run_if: "true"
+inputs:
+- PROJECT_SLACK_CHANNEL: '#bitrise'
+  opts:
+    title: Default channel for outputting succeeded build info, links etc
+- ERROR_SLACK_CHANNEL: '#bitrise'
+  opts:
+    title: Default channel for outputting step errors
+- HOCKEYBUILDSJSON: ""
+  opts:
+    description: |-
+      The build information format is:
+      ```
+      [
+        {
+          "build": "app-first.apk",
+          "hockeyId": "11111111111111111111111111111",
+          "appId": "dk.nodes.citestflavors.firstskin","mappingFile": "null"
+        },
+        {
+          "build": "app-second.apk",
+          "hockeyId": "11111111111111111111111111111",
+          "appId": "dk.nodes.citestflavors.secondskin","mappingFile": "null"
+        }
+      ]
+      ```
+    title: JSON String containing build info


viktorbenei:
Hi @johsoe ,

Thanks for the PR! ;)

After a quick glance on the step/description, could you please tell us how this step is different compared to our Hockey deploy steps?

johsoe:
A quick note on the reasons why we created this:

We are an agency which develops apps for many clients and more and more of these clients are buying reskins of their apps. This means we end up with many many different builds, which are mapped to different hockey ids.

The idea of this step is to build whatever is needed via gradle/some upcoming iOS build stuff, output some json in the form:

```
[
	{
		"build": "app-first.apk",
		"hockeyId": "11111111111111111111111111111",
		"appId": "dk.nodes.citestflavors.firstskin","mappingFile": "null"
	},
	{
		"build": "app-second.apk",
		"hockeyId": "11111111111111111111111111111",
		"appId": "dk.nodes.citestflavors.secondskin","mappingFile": "null"
	}
]
```

johsoe:
@viktorbenei Was already writing a comment on that, see above ^

johsoe:
We had a private CI toolchain on Gitlab CI, that we are working on moving to Bitrise, so expect more to come :)

viktorbenei:
Sounds good :)

As this is a new step I'll wait for a second review ( @godrei / @trapacska ), but I think it should be OK to have this as an additional step for Hockey.

> This step then iterates over the build elements, uploads them, and summarizes this into slack message containing either errors or links to all the builds

Sounds awesome btw ;) Can it handle both iOS and Android deploys?

> We had a private CI toolchain on Gitlab CI, that we are working on moving to Bitrise, so expect more to come :)

If you have the time feel free to ping us through the onsite chat on bitrise.io, would love to hear more about this ;)

johsoe:
The plan is to have a platform agnostic json format that will upload whatever, so in theory it's going to support whatever HockeyApp accepts :)

viktorbenei:
Sounds awesome! Just thinking, would you be open to merge the steps? If not that's not a problem either, but having one step which could handle both the standard single input flow as well as a more complex one like your step, sounds tempting 😉

johsoe:
It would make sense long term, but it's pretty crucial for us at the moment, since we have that need right now.

We might also need to align on feature sets, we have some pretty specific logging requirements (due to PM's wanting statuses, build links etc). That could be toggled off via input flags or something, I guess.

viktorbenei:
@johsoe sure, definitely makes sense. In that case my question would be: do you need this step to be shared in the StepLib? You can easily use the step without sharing it, using the `git::` style step referencing (http://devcenter.bitrise.io/bitrise-cli/steps/#special-step-sources). There's absolutely no requirement for a step to be shared.

The only downside if it's not shared in the StepLib is that the step's UI (editable input fields, icon, ..) won't be shown in the workflow editor on bitrise.io (only a placeholder is presented on the UI), but you can of course still configure it in yml mode there and use it without any limitation. Additionally if you run the workflow editor locally on your Mac/Linux that supports custom sources including the `git::` steps, with full UI support (https://discuss.bitrise.io/t/workflow-editor-v2-open-source-offline-workflow-editor/39).

In general we only accept steps in this StepLib which are generic enough to be useful for others too; if you have a custom step tailored for your own needs it's usually better to just use it directly with a `git::` ref so that you don't even have to wait for a review from us.

Just to be clear, I'm not against accepting this step at all, just wanted to make sure you know about the alternatives which might fit your need better based on what you wrote :)

johsoe:
I understand, so I agree that it makes sense to merge the two and then use the direct git ref to our step in the mean time. We just wanted the fancy icons etc :)

We don't have a lot of expertise in Go though, which is why we used Ruby - do you have strong opinions of Ruby vs Go?

viktorbenei:
> We just wanted the fancy icons etc :)

Yeah, those are tempting :D We should make it possible to define an icon for custom steps..

> We don't have a lot of expertise in Go though, which is why we used Ruby - do you have strong opinions of Ruby vs Go?

Well, to be honest we do have a strong preference towards Go. Back when we started we wrote most of our steps in Bash and Ruby; simple ones in Bash more complex ones in Ruby. But over the time we had so many issues with Ruby (mainly gems/dependencies and surprising incompatibility between minor version changes, e.g. Ruby 2.4 broke almost all of the popular CLI tools...) that we slowly started to rewrite those in Go.

The thing is, if you have 100+ steps/repos to maintain Go shines with it's legendary backwards compatibility and with the simplicity of being able to have all the dependencies committed into the repository (which means that there's absolutely no dependency on 3rd party services; if the git clone of the step repo is successful you have everything there to be able to run/build the step).

Go is also a super simple language. It's not inspiring, it's quite boring actually but that's the point, it gets out of your way and it does it in a predictable and reliable way.

godrei:
Hi @johsoe sorry for my late response and thanks for the new step!

I'm not sure if we should merge this step, because of:

- our build steps are not prepared to generate `HOCKEYBUILDSJSON` and most of them generates single artifacts (expect gradle-runner in some cases), so currently our build steps can not feed your step's  input.
- this step combines and extends two of our currently living steps (hockeyapp-deploy and slack), we designed steps to be an independent building blocks of the bitrise automations

As @viktorbenei mentioned our steplib contains generic steps, which are useful for other users as well. I think your step is specialised for your requirements,  you can use your step using the `git::` style step referencing (http://devcenter.bitrise.io/bitrise-cli/steps/#special-step-sources)

If you believe I did not understand what this step does and you think this could help others too please let me know!

viktorbenei:
I have to agree with @godrei here, the step seems to be too specialized to be useful for others, closing this PR.

That said we talked quite a bit about this, how custom steps could be handled better. So far what we thought:

1. Make it possible to use get a UI for `git::` style referenced steps on bitrise.io. Right now this works in local but not on bitrise.io; on bitrise.io you'll get a "placeholder" step on the UI where you can't set the inputs, only in `bitrise.yml` mode you can.
1. Make it simple to add custom steps, by just providing the git clone URL in the editor. Something like: click on the `+` (add step) button, and in the right sidebar there should be an option to "add a custom step via git clone URL" or similar.
1. Create an alternative "community step collection" without restrictions. There every step would be listed with a git clone URL, so if you like it you could simply just add it in the editor with the "custom step via git clone URL" option. Of course you could also just highlight this in the step's README, a simple how-to/section like "add it in bitrise.yml with `git::...` or in the editor with the git clone URL: ...".

What do you think @johsoe ?

johsoe:
@viktorbenei Totally understandable, we ended up just making it "git::..." step. Our step is pretty specific to our use case.

We might add our functionality as a PR in the Go Hockey step at some point though.

viktorbenei:
Awesome, let us know if we can help with anything! ;)

P.S.: also scheduled a couple of improvements for the workflow editor to make this use case (custom step with git:: ref) easier everywhere :)

PR #1068, opened 2017-06-20
move-or-copy-files 0.0.1

Dave005:

### DIFF
--- a/steps/move-or-copy-files/0.0.1/step.yml
+++ b/steps/move-or-copy-files/0.0.1/step.yml
@@ -1,0 +1,56 @@
+title: Move or Copy Files
+summary: |
+  Moves or copies files
+description: |
+  -
+website: https://github.com/...
+source_code_url: https://github.com/...
+support_url: https://github.com/.../issues
+published_at: 2017-06-20T10:20:33.1127341+02:00
+source:
+  git: https://github.com/Dave005/bitrise-step-move-or-copy-files.git
+  commit: 1618f8204b687a0914de1d5f1f849ee60f1bfc43
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+type_tags:
+- utility
+toolkit:
+  bash:
+    entry_file: step.sh
+deps:
+  brew:
+  - name: git
+  - name: wget
+  apt_get:
+  - name: git
+  - name: wget
+is_requires_admin_user: true
+is_always_run: false
+is_skippable: false
+run_if: ""
+inputs:
+- input_file: null
+  opts:
+    description: |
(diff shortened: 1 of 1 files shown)


steps/move-or-copy-files/0.0.1/step.yml
```diff
@@ -0,0 +1,56 @@
+title: Move or Copy Files
+summary: |
+  Moves or copies files
+description: |
+  -
```
viktorbenei: please fill this out

steps/move-or-copy-files/0.0.1/step.yml
```diff
@@ -0,0 +1,56 @@
...
+website: https://github.com/...
```
viktorbenei: please fill these URLs too, can be simply the github page of the repo (`https://github.com/Dave005/bitrise-step-move-or-copy-files`)

steps/move-or-copy-files/0.0.1/step.yml
```diff
@@ -0,0 +1,56 @@
...
+  commit: 1618f8204b687a0914de1d5f1f849ee60f1bfc43
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+type_tags:
+- utility
+toolkit:
+  bash:
+    entry_file: step.sh
+deps:
```
viktorbenei: please remove this (the whole `deps`) if you don't need any dependency

steps/move-or-copy-files/0.0.1/step.yml
```diff
@@ -0,0 +1,56 @@
...
+    summary: ""
+    title: Source File
+- opts:
+    description: |
+      Path to the target file
+    is_required: true
+    summary: ""
+    title: Target File
+  output_file: null
+- move_action: false
```
viktorbenei: don't use bool, as when a bool is serialized into string (env vars are always strings) the string representation might depend on the language. E.g. might be serialized as "true/false" or "True/False" or even something else like "Yes/No"

viktorbenei: simply guarding the `false` with double quotes is fine (`"false"`), but in this case I'd say it'd be even better to have the options specified, e.g. instead of `"true"` and `"false"` could be `"move"` and `"copy"`

viktorbenei:
Hi @Dave005 and thanks for the Step! ;)

The step's code seems 👍  but please fix the things I listed above. Important: please do NOT remove the existing tag/version, simply do the changes and create a new version of the step (e.g. 0.0.2) and then submit that version in a new PR

If you have any questions just let us know!

PR #1101, opened 2017-07-10
Openstf disconnect 0.0.1

koral--:


godrei:
Hi @koral-- ,
your __Openstf disconnect__ step looks great to me. 
I have only one question: are you sure about the `type_tags: test`? I think it should be `utility`, since it does not run test directly, just connects to your device farm.
What do you think?

viktorbenei:
I'd vote for `test`, that's where **I** would look for this step I think.

koral--:
Hmm, this step only *dis*connects/releases devices acquired previously using Open STF Connect step.
Indeed, literally, it does not perform any tests on its own. But there is no usage of test farm other than tests :)
So I chosen test tag for both connect and disconnect steps.

If `test` tag should be used only for steps which perform actual tests then also steps like `aws-device-farm-file-deploy` should be recategorized to `utility`, shouldn't they?

viktorbenei:
I agree, for me it seems reasonable to use the `test` step for these steps, as I mentioned above ;)

viktorbenei:
That said there's an issue with this PR - it includes both the **connect** and the **disconnect** steps. Every PR should only include changes for only one step.

How did you share this step @koral-- ? Did you use `bitrise run share-this-step` or did you use a manual approach?

koral--:
I first used bitrise plugin then added icon manually in next commit.
In fact I included only one step in single PR, namely I repeated bitrise share and icon adding for the second step but due to the fact the branch of 2nd step is based on one from first one, GitHub UI list included also commits from first one.

viktorbenei:
> due to the fact the branch of 2nd step is based on one from first one

the `bitrise run` / `stepman share` should always start from the **master** branch of your steplib fork, not from your last branch

can it be that the step ID in one of your steps' `bitrise.yml` was the same as other? if the ID was the same then this might happen

koral--:
> the bitrise run / stepman share should always start from the master branch of your steplib fork, not from your last branch

I rebased that branch on forked repo before adding a second icon.
BTW in cases (like this) of sharing multiple steps in a row is it preferred to create separate PRs based on master for each one and then update subsequent PRs by rebasing after previous one are merged?

I reshared this step by rebasing and force pushing to `openstf-disconnect-0.0.1` and now there only commits related to single step. Is it OK or should I close this PR and create new one?

viktorbenei:
> Is it OK or should I close this PR and create new one?

Rebase is fine, the PR is 👍 now ;)

> I rebased that branch on forked repo before adding a second icon.

Ahh, I see. We usually don't rebase our forks at all; unless you need that to be up to date there really is no reason to do that. In general all we do is:

1. at the very first time we fork this steplib repo
2. from that point, we just `bitrise run share-this-step` for steps

No rebase/update; that way the PRs only include the diffs for that specific step/change.

godrei:
@koral-- thanks for the new step! 👍

PR #1200, opened 2017-09-29
Update xcode-test docs

Mazyod:
-derivedData $DERIVED_DATA isn't actually being added to the default `xcodebuild` command, which was a surprise for me when I added `DERIVED_DATA` as an env variable, but the actual path didn't change.


godrei:
Hi @Mazyod,
thanks for the PR!
The problem is that, this fix should be created in the step repo and release a new version of the step, to properly apply this fix.
If you only create the fix here, at the step definition, the issue will remain in the step repo and when we release a new version of the step, the issue will come up again.
Please create the same fix at this repo: https://github.com/bitrise-io/steps-xcode-test

PR #1722, opened 2018-09-05
revision + 2 new handlers added

trapacska:

### DIFF
--- a/handlers/validate/steplib/steplib.go
+++ b/handlers/validate/steplib/steplib.go
@@ -1,0 +1,116 @@
+package steplib
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/go-utils/pathutil"
+	version "github.com/hashicorp/go-version"
+)
+
+type StepLib struct {
+	Steps []Step
+}
+
+func (sl StepLib) FindStepByID(id string) Step {
+	for _, step := range sl.Steps {
+		if step.ID == id {
+			return step
+		}
+	}
+	return Step{}
+}
+
+func NewStepLib(rootPath string) (StepLib, error) {
+	stepLib := StepLib{}
+	stepsDir := filepath.Join(rootPath, "steps")
+	stepDirs, err := ioutil.ReadDir(stepsDir)
+	if err != nil {
+		return StepLib{}, nil
+	}
+	for _, stepDirInfo := range stepDirs {
+		if !stepDirInfo.IsDir() {
+			continue
+		}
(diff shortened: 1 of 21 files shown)


handlers/validate/steplib/steplib.go
```diff
@@ -0,0 +1,116 @@
...
+			versionFound = v
+		}
+	}
+	if latestVersion.String() == "0.0.0" {
+		return Version{}, fmt.Errorf("unable to find latest version")
+	}
+	return versionFound, nil
+}
+
+func findPreviousVersion(versions []Version, latest Version) (Version, error) {
```
godrei: unused function

handlers/validate/steplib/steplib.go
```diff
@@ -0,0 +1,116 @@
...
+	"path/filepath"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/go-utils/pathutil"
+	version "github.com/hashicorp/go-version"
+)
+
+type StepLib struct {
```
godrei: Why do not you use the stepman models? both the stepli, step and steplib parsing is already implemented there

handlers/validate/steplib/version.go
```diff
@@ -0,0 +1,22 @@
+package steplib
+
+import (
+	stepmanModels "github.com/bitrise-io/stepman/models"
+)
+
+type Version struct {
+	ID          string
+	Previous    *Version
```
godrei: unused property

handlers/validate/validators/valueoptions/valueoptions.go
```diff
@@ -105,16 +57,10 @@ func (v *Validator) Validate() error {
 	return nil
 }
 
-func validateStepYML(vers, stepID string) error {
-	path := filepath.Join("steps", stepID, vers, "step.yml")
+// - %s@%s has invalid value_options
```
godrei: ?

handlers/validate/steplib/steplib.go
trapacska: No significant difference between the this and the stepman model, however I needed the step as an array rather than a hash. (Using ID in the key and the struct field Id as well would be duplicate)

handlers/validate/steplib/version.go
trapacska: It is a preparation for the next validator task(diff the whole code change(not only the step.yml) and find important changes)

godrei: could you please link where do you use this property?

trapacska: I started to develop this feature 2 weeks ago here: https://github.com/bitrise-io/bitrise-steplib/tree/diff-tool and it is not yet updated to use the new StepLib model. (Which is in this PR)

PR #1749, opened 2018-09-26
bitrise-to-appscan 0.2.0

ipiranga-atq:
TagCheck

- [*] __I will not move an already shared step version's tag to another commit__
- [*] I read the [Step Development Guideline](https://github.com/bitrise-io/bitrise/blob/master/_docs/step-development-guideline.md)
- [*] I have a test for my Step, which can be run with `bitrise run test` (in the step's repository)
- [*] I did run `bitrise run audit-this-step` (in the step's repository - note, if you don't have this workflow in your `bitrise.yml`, [you can copy it from the step template](https://github.com/bitrise-steplib/step-template/blob/master/bitrise.yml).)
- [*] I read and accept the [Abandoned Step policy](https://github.com/bitrise-io/bitrise-steplib#abandoned-step-policy)

### DIFF
--- a/steps/bitrise-to-appscan/0.2.0/step.yml
+++ b/steps/bitrise-to-appscan/0.2.0/step.yml
@@ -1,0 +1,72 @@
+title: AppScan Security Analysis
+summary: Performs security analysis using the IBM AppScan Cloud.
+description: |
+  Performs security analysis using the IBM AppScan Cloud.
+website: https://github.com/ipiranga-atq/bitrise_appscan_step
+source_code_url: https://github.com/ipiranga-atq/bitrise_appscan_step
+support_url: https://github.com/ipiranga-atq/bitrise_appscan_step
+published_at: 2018-09-26T15:58:54.306065087-03:00
+source:
+  git: https://github.com/ipiranga-atq/bitrise_appscan_step.git
+  commit: 96515265be0c324bcd00d1ad96a83b2a6fe2677d
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+type_tags:
+- utility
+toolkit:
+  go:
+    package_name: github.com/ipiranga-atq/bitrise_appscan_step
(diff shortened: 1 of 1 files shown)


steps/bitrise-to-appscan/0.2.0/step.yml
```diff
@@ -0,0 +1,72 @@
...
+run_if: ""
+inputs:
+- app_path: $BITRISE_APK_PATH
+  opts:
+    description: |
+      Type of application's installer (APK or IPA)
+    is_expand: true
+    is_required: true
+    title: 'Choose file type: APK/IPA'
+    value_options:
```
godrei: i see your point but it is a bit strange to use `BITRISE_APK_PATH` and `BITRISE_IPA_PATH`.
Value Options means that you need to choose from the listed options, but what if someone uses his own script to build an ipa/apk file and uses different env var name?

I would recommend mentioning that `BITRISE_APK_PATH` or `BITRISE_IPA_PATH` can be used to fill this value, but i would ask you remove the `value_options ` from this input.

steps/bitrise-to-appscan/0.2.0/step.yml
```diff
@@ -0,0 +1,72 @@
...
+- opts:
+    is_expand: true
+    is_required: true
+    title: Username for AppScan
+  user_name: null
+- opts:
+    is_expand: true
+    is_required: true
+    title: Password for AppScan
+  user_password: null
```
godrei: you can mark inputs as `is_sensitive` like: https://github.com/bitrise-io/steps-deploy-to-bitrise-io/blob/master/step.yml#L68
if you set this property on an input, our ui will force the users to set secret env var value for this inputs, which are auto redacted from our build logs: https://blog.bitrise.io/we-will-protect-your-secrets-with-the-new-sensitive-inputs

PR #1893, opened 2019-02-01
telegram-notify 1.0.5

mcaldas:
TagCheck

### DIFF
--- a/steps/telegram-notify/1.0.5/step.yml
+++ b/steps/telegram-notify/1.0.5/step.yml
@@ -1,0 +1,61 @@
+title: Send Telegram Message
+summary: |
+  Notify telegram group or chat on builds status.
+description: |
+  This step will send a message to the provided [telegram](https://core.telegram.org/bots) chat.
+  Includes the build status and download URL. You can also add a custom message to be sent with the notification.
+  This step always runs, no matter if build succeded or failed.
+website: https://github.com/mcaldas/bitrise-step-telegram-notify
+source_code_url: https://github.com/mcaldas/bitrise-step-telegram-notify
+support_url: https://github.com/mcaldas/bitrise-step-telegram-notify/issues
+published_at: 2019-02-02T00:27:06.060346+02:00
+source:
+  git: https://github.com/mcaldas/bitrise-step-telegram-notify.git
+  commit: a34bfeb6c68e4b33e42e67242bb91a729ca5f373
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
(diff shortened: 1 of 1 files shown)


steps/telegram-notify/1.0.5/step.yml
```diff
@@ -0,0 +1,61 @@
...
+- osx-10.10
+- ubuntu-16.04
+type_tags:
+- notification
+toolkit:
+  bash:
+    entry_file: step.sh
+deps:
+  brew:
+  - name: curl
```
lszucs: `curl` is readily available on our stacks. Declaring it will cause the builds to run longer.

For a full list of installed tools, see: https://github.com/bitrise-io/bitrise.io/tree/master/system_reports

mcaldas: Ok, I will remove that then, right?

mcaldas: @lszucs another question, is the approach on taking the input variables (not the env variable) using its name the right one?
Because trying locally it's not working, but not sure if its because of the `bitrise.secrets.yml`?
Im talking about for example, getting the actual value set in the "telegram chat id" field as `$telegram_chat_id` in the bash script

mcaldas: @lszucs ok, so u keep me posted about it?
Regarding the `.bitrise.secrets.yml` yes, thats the file I used, but for some reason it didnt expand those variables, only the regular env variables. Would u mind taking a look at the `.sh` and confirm its correct?

lszucs: Hi @mcaldas,

Sure, I'll check it out today. I will get back to you later.

The guideline regarding `deps` is to list the runtime dependencies only.

For example: for a `toolkit: go` step, you would not list Go, because the step wouldn't be running in the first place without it, but you would list all external commands you would call: git, wget, whatever, so that if you run your step outside our environments, the Bitrise CLI can prepare them. The CLI is not going to reinstall existing tools if they are listed there.

So, in your case, you can leave `curl` there, if your script actually calls it.

mcaldas:
Cool, so yes it calls curl.

Let me know about the variables and if it’s ok feel free to merge it.

steps/telegram-notify/1.0.5/step.yml
mcaldas: @lszucs any update on this?

trapacska:
Thank your for the PR! LGTM 🚀

PR #1894, opened 2019-02-04
android-detekt 0.0.2

dzinek:
TagCheck

Second version from PR #1868 

### DIFF
--- a/steps/android-detekt/0.0.2/step.yml
+++ b/steps/android-detekt/0.0.2/step.yml
@@ -1,0 +1,74 @@
+title: Android Detekt
+summary: |
+  This step runs detekt gradle task on your project.
+description: |
+  This step runs detekt gradle task on your project.
+website: https://github.com/dzinek/bitrise-step-android-detekt
+source_code_url: https://github.com/dzinek/bitrise-step-android-detekt
+support_url: https://github.com/dzinek/bitrise-step-android-detekt/issues
+published_at: 2019-02-04T06:52:05.394754+01:00
+source:
+  git: https://github.com/dzinek/bitrise-step-android-detekt.git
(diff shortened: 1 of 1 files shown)


steps/android-detekt/0.0.2/step.yml
```diff
@@ -0,0 +1,74 @@
...
+project_type_tags:
+- android
+type_tags:
+- test
+toolkit:
+  go:
+    package_name: github.com/dzinek/bitrise-step-android-detekt
+deps:
+  apt_get:
+  - name: git
```
lszucs: `git` dependency remains.

Sorry, but I'll have to decline. :(

dzinek: @lszucs 
Git dependency remains and from my perspective it should remain, at least until I saw your comment...

I have removed all dependencies from brew as not needed - it is not a MacOS nor iOS build step.
I have removed also wget from apt-get part as it is not needed.
But I was pretty sure that I still need git dependency (removed from brew so it is not duplicated) - once I need it eg. to build this step. Now I suppose that I was wrong - I don't need it during real step running.

lszucs: Hi @dzinek, 

There was a misunderstanding on our side regarding these dependencies. Please do not close this PR, until I get to you with the official guideline. My apologies for the inconvenience.

lszucs: Hello @dzinek,

The guideline regarding `deps` is to list the runtime dependencies.

For example: for a `toolkit: go` step, you would not list Go, because the step wouldn't be running in the first place without it, but you would list all external commands you would call: git, wget, whatever, so that if you run your step outside our environments, the Bitrise CLI can prepare them.

I shall approve your PR.

PR #1909, opened 2019-02-19
Additional collection check for older CLI versions

trapacska:


handlers/validate/validators/collection/collection.go
```diff
@@ -27,12 +36,100 @@ func (v *Validator) IsSkippable() bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := resp.Body.Close(); err != nil {
```
koral--: Shouldn't closing be deferred before trying to read body? E.g. like that: https://github.com/bitrise-io/bitrise/blob/master/cli/update.go#L159

trapacska: I think if there is an error reading resp.Body in line 51 then that is a case when you won't need to close the connection, and if there was no error then I close it manually after reading it here because I know that I won't need that connection anymore. AFAIK if I read from a connection until EOF and the connection type is not keep-alive then it will be closed automatically by the server. Please correct me if I am wrong and I will place it inside of a defer instantly. 🙂

viktorbenei: I think @koral-- is right, or at least that is the recommended approach and the safest one: https://golang.org/pkg/net/http/

Basically unless you get an `err` at `resp, err := http.DefaultClient.Do(req)` you should `defer resp.Body.Close()`, as it won't cause any issue in any case but not calling it at a case where you should can cause memory leak.

trapacska: I see. Thanks @viktorbenei and @koral-- !

PR #1926, opened 2019-03-07
Unity Install 0.1.0

bdrelling:
TagCheck

### DIFF
--- a/steps/unity-install/0.1.0/step.yml
+++ b/steps/unity-install/0.1.0/step.yml
@@ -6,7 +6,7 @@ description: |
 website: https://github.com/bdrelling/bitrise-step-unity-install
 source_code_url: https://github.com/bdrelling/bitrise-step-unity-install
 support_url: https://github.com/bdrelling/bitrise-step-unity-install/issues
-published_at: 2019-03-04T19:48:11.697942-06:00
+published_at: 2019-03-06T20:47:18.791348-06:00
 source:
   git: https://bdrelling@github.com/bdrelling/bitrise-step-unity-install.git
   commit: 63ce02bbaf1443921cd44f382cbf3feac9647875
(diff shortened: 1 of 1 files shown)


lszucs:
Hi @bdrelling,

Thanks for the PR! Unfortunately, there are some issues to be sorted out.

In order to have your step imported, you will have to close this PR and open a new one, because you have moved the tag (pushed commits and retagged), after opening it.

But don't close it yet, I'll tell you why. If you fix the errors and update the tag, I can rerun our automatic validations and we'll create a new PR only when all is good.

You will have to fix the following build issues:

```
Running validations:
- (1/4) Running: Ensure CLI is working properly with the new collection
 > Successful
- (2/4) Running: Ensure all value_option is string type
 > Failed, error:
   - type(bool) in value_options with value: true at input: Build iOS?
   - type(bool) in value_options with value: false at input: Build iOS?
   - type(bool) in value_options with value: true at input: Build Android?
   - type(bool) in value_options with value: false at input: Build Android?
```

The mentioned options must be specified as string types in the step.yml.

Once done, push the fixes and update the `0.1.0` tag and I'll rerun the build. :)

bdrelling:
@lszucs I'm actually not sure how this PR has a reference to a previous commit. The tag I have in my repo is where it should be on `274f253`: https://github.com/bdrelling/bitrise-step-unity-install/commits/master

What happens if I run the `bitrise share` command again in the same repo for the same tag? Does it update the tag, or leave the tag where it was?

I think if I want that commit to be the one used, I should close this PR again though either way, right?

trapacska:
@bdrelling 
> Why does that throw an error? Does it need to explicitly say "true" even when listed?

Yes, those must be string. Like you mentioned: `"true"` for example.

After the modifications you need to re-share the step.

Previously when you shared the step the tag 0.1.0 was at commit hash:
Screen Shot 2019-03-18 at 12 35 30 PM
But now it is moved moved it to the:
Screen Shot 2019-03-18 at 12 35 55 PM
commit hash.

Please when you are done with the changes, create a new version share and do not move any tags.

Thanks! 🚀

PR #1929, opened 2019-03-10
swift-package-manager-test-for-mac 0.0.1

kitasuke:
TagCheck

### DIFF
--- a/steps/swift-package-manager-test-for-mac/0.0.1/step.yml
+++ b/steps/swift-package-manager-test-for-mac/0.0.1/step.yml
@@ -1,0 +1,62 @@
+title: Swift Package Manager Test for Mac
+summary: |
+  Runs Test on your macOS project with Swift Package Manager
+description: |
+  Runs Swift Package Manager's `swift test` action for macOS app projects.
+website: https://github.com/kitasuke/bitrise-step-swift-package-manager-test-for-mac
+source_code_url: https://github.com/kitasuke/bitrise-step-swift-package-manager-test-for-mac
+support_url: https://github.com/kitasuke/bitrise-step-swift-package-manager-test-for-mac/issues
+published_at: 2019-03-10T16:07:57.802577+09:00
+source:
+  git: https://github.com/kitasuke/bitrise-step-swift-package-manager-test-for-mac.git
+  commit: b7a1e866f8af3402c6a4c9cd9285b4a31b923be6
+host_os_tags:
+- osx-10.10
+project_type_tags:
+- macos
+type_tags:
+- test
+toolkit:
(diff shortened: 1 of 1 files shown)


steps/swift-package-manager-test-for-mac/0.0.1/step.yml
```diff
@@ -0,0 +1,62 @@
...
+      Should run the tests in parallel?
+    title: Parallel test
+    value_options:
+    - "yes"
+    - "no"
+outputs:
+- BITRISE_SWIFT_TEST_RESULT: null
+  opts:
+    title: Result of the tests. 'succeeded' or 'failed'.
+    value_options:
```
koral--: `value_options` is meaningful only for inputs

steps/swift-package-manager-test-for-mac/0.0.1/step.yml
```diff
@@ -0,0 +1,62 @@
...
+- is_parallel: "no"
+  opts:
+    description: |
+      Should run the tests in parallel?
+    title: Parallel test
+    value_options:
+    - "yes"
+    - "no"
+outputs:
+- BITRISE_SWIFT_TEST_RESULT: null
```
koral--: What is the purpose of this output?
Description suggests that the only purpose of this step is performing tests, so if they fail then step execution will just fail. No need for additional variable.

kitasuke: That's absolutely right. Will fix it

kitasuke: Addressed your feedback

steps/swift-package-manager-test-for-mac/0.0.1/step.yml
```diff
@@ -0,0 +1,54 @@
...
+is_always_run: false
+is_skippable: false
+run_if: ""
+inputs:
+- build_path: ./.build
+  opts:
+    description: |
+      A build path, relative to the working directory.
+    is_expand: true
+    is_required: false
```
trapacska: @kitasuke why is it not required? Can the step work without setting build_path?

kitasuke: Can't for the step, but can work for `swift test` since default value is set for build_path. Do you think we should make it required for the case?

kitasuke: I now feel like I should... Will fix it.

kitasuke: Addressed

PR #1942, opened 2019-03-19
Unity Install 0.1.0

bdrelling:
TagCheck

Two Notes:
- Was having issues with local git install on Mojave. I think it's resolved now, so you shouldn't see any tag moves or otherwise.
- `bitrise run test` works, but I don't know if it's an optimal test. Would like some input, or I could tweak it over time as needs change.

### DIFF
--- a/steps/unity-install/0.1.0/step.yml
+++ b/steps/unity-install/0.1.0/step.yml
@@ -1,0 +1,92 @@
+title: Unity Install
+summary: |
+  Downloads and installs Unity packages
+description: |
+  Downloads and installs Unity packages
+website: https://github.com/bdrelling/bitrise-step-unity-install
+source_code_url: https://github.com/bdrelling/bitrise-step-unity-install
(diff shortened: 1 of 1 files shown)


steps/unity-install/0.1.0/step.yml
```diff
@@ -0,0 +1,92 @@
...
+      where `a220877bc173` is the ID for v2018.3.6.  \nExample **What's new?** page:
+      https://unity3d.com/unity/whats-new/2018.3.6\n"
+    is_expand: true
+    is_required: false
+    summary: The ID of the installer. If no value is specified, it will use the latest
+      stable release version of Unity.
+    title: Unity Installer ID
+- opts:
+    description: |
+      Downloads and installs Unity Android Target Support packages.
```
trapacska: Here you mention Android instead of iOS

bdrelling: Thanks, good catch.

PR #1946, opened 2019-03-21
Unity Install 0.1.0

bdrelling:
TagCheck

### DIFF
--- a/steps/unity-install/0.1.0/step.yml
+++ b/steps/unity-install/0.1.0/step.yml
@@ -1,0 +1,92 @@
+title: Unity Install
+summary: |
+  Downloads and installs Unity packages
+description: |
+  Downloads and installs Unity packages
+website: https://github.com/bdrelling/bitrise-step-unity-install
+source_code_url: https://github.com/bdrelling/bitrise-step-unity-install
+support_url: https://github.com/bdrelling/bitrise-step-unity-install/issues
+published_at: 2019-03-21T11:28:54.545386-05:00
+source:
+  git: https://github.com/bdrelling/bitrise-step-unity-install.git
+  commit: a168322102969eb88d649d0d3304e756da26df9c
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+project_type_tags:
+- xamarin
+type_tags:
+- installer
+is_requires_admin_user: true
+is_always_run: false
+is_skippable: false
+run_if: ""
+inputs:
+- installer_id: ""
+  opts:
+    description: "The installer ID can be found within the Component Installer URLs
+      on the \"What's new?\" page for the Unity version you're trying to install.
+      If no value is specified, it will use the latest stable release version of Unity.
+      \ \nExample Component Installer URL: `https://download.unity3d.com/download_unity/a220877bc173/TargetSupportInstaller/...`
+      where `a220877bc173` is the ID for v2018.3.6.  \nExample **What's new?** page:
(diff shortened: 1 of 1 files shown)


steps/unity-install/0.1.0/step.yml
```diff
@@ -0,0 +1,92 @@
...
+support_url: https://github.com/bdrelling/bitrise-step-unity-install/issues
+published_at: 2019-03-21T11:28:54.545386-05:00
+source:
+  git: https://github.com/bdrelling/bitrise-step-unity-install.git
+  commit: a168322102969eb88d649d0d3304e756da26df9c
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+project_type_tags:
+- xamarin
```
BirmacherAkos: Is there a reason for it supports Xamarin only?

What about:
- ios
- android
- react-native

steps/unity-install/0.1.0/step.yml
```diff
@@ -0,0 +1,92 @@
...
+type_tags:
+- installer
```
BirmacherAkos: I think it would be a good idea to add `utility` next to the installer type_tag:

```
type_tags:
- installer
- utility
```

steps/unity-install/0.1.0/step.yml
bdrelling: My understanding is that the Xamarin stack will have the necessary components to reduce overhead from installation. Is this not the case? Honestly, there should be a Unity project type in the future to replace this with. I would not think `react-native` should work at all, and `ios` and `android` won't include the `dotnet` SDKs.

steps/unity-install/0.1.0/step.yml
bdrelling: Can you explain why so I understand?

bdrelling:
I've been trying to get this step through the PR process a number of items. Every time, there are new comments/suggested changes, and I'm only changing stuff from the previous PR. Is it possible to get a more thorough review so I can address these things all at once? This is becoming quite a process otherwise.

Thanks.

steps/unity-install/0.1.0/step.yml
lpusok: That's right.

godrei:
Hi @bdrelling 
sorry for the inconvenience, a StepLib PR review should be much smoother. We have a ready to sprint task to explain the StepLib PR requirements in more details.

We reviewed this step and the only remaining thing is the `project_type_tags` update. The installer type tag is fine.

Our scanner identifies the project type and stores in the generated bitrise config (bitrise.yml). The project type is used by the workflow editor, to filter the matching steps to the project.

Since we do not provide scanner and steps for Unity, for now, you should leave the `project_type_tags` empty to make it visible for every project type by default.

Unity support is on our roadmap so the project type support will come soon.

lpusok:
Closing PR.

> Hi @bdrelling
> sorry for the inconvenience, a StepLib PR review should be much smoother. We have a ready to sprint task to explain the StepLib PR requirements in more details.
> 
> We reviewed this step and the only remaining thing is the `project_type_tags` update. The installer type tag is fine.
> 
> Our scanner identifies the project type and stores in the generated bitrise config (bitrise.yml). The project type is used by the workflow editor, to filter the matching steps to the project.
> 
> Since we do not provide scanner and steps for Unity, for now, you should leave the `project_type_tags` empty to make it visible for every project type by default.
> 
> Unity support is on our roadmap so the project type support will come soon.

PR #2018, opened 2019-06-14
applivery-ios 0.3.1

CesarTrigo:
TagCheck

### Added
- Support to `BITRISE_CERTIFICATE_URL`
- Support to `BITRISE_CERTIFICATE_PASSPHRASE`

### Fixed
- Wrong `BITRISE_IPA_PATH` making it impossible to find a valid `.ipa` file to deploy

### DIFF
--- a/steps/applivery-ios/0.3.1/step.yml
+++ b/steps/applivery-ios/0.3.1/step.yml
@@ -1,0 +1,94 @@
+title: Applivery.com v3 iOS Deploy
+summary: Deploy your awesome iOS App to Applivery.com
+description: |-
+  Deploy an iOS application to [Applivery](http://www.applivery.com),
+  add notes and notify Collaborators and Employees.
+
+  Register a Applivery account at [http://www.applivery.com/](http://www.applivery.com)
+  and create an App to utilize this step.
+
+  You also need to get your *App Token* of your App that can be found under your App Settings.
+website: https://github.com/applivery/steps-applivery-ios-deploy
+source_code_url: https://github.com/applivery/steps-applivery-ios-deploy
+support_url: https://github.com/applivery/steps-applivery-ios-deploy/issues
+published_at: 2019-06-14T12:44:16.708514+02:00
+source:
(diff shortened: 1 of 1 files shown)


CesarTrigo:
Hi @BirmacherAkos 
We've been forced to submit new versions of our steps due to a **recent change in the new version of our platform**. We also have taken the opportunity to fine tune based on your suggested and updated both icons.

Related PR are:
- [iOS Step v0.3.1](https://github.com/bitrise-io/bitrise-steplib/pull/2018)
- [Android Step  v0.3.1](https://github.com/bitrise-io/bitrise-steplib/pull/2016)
- [New icons](https://github.com/bitrise-io/bitrise-steplib/pull/2017)

**Any possibility to speed up the validation / merge will be more than welcome**

Best,
C.

BirmacherAkos:
Hi

I checked your changes in your step, and have some questions:

1) You want to upload the provisioning profiles, certificates and it's passphrases to Applivery. Can I ask why?
2) This could be a little risky. I would suggest adding another step input with a default false value which could enable the upload.
So if the user wants to upload their code signing files, they could enable it.
Something like this:

```
- upload_code_signing: "false"
    opts:
      title: Enable to upload certificates and provisioning profiles to Applivery.
      summary: Applivery will use these files to...
      description: |
	If you enable this option the step will download your code signing files from Bitrise Code Signing Tab and it will upload it to...
      is_required: true
      value_options:
        - "true"
        - "false"
```

CesarTrigo:
Hi @BirmacherAkos,
Thank you for your review, comments and suggestions.
I do agree with you that it would be better to not upload them directly but ask the user first. While we adapt the integration to do so. I'll close this PR and perform a new one without including `BITRISE_CERTIFICATE_URL ` and `BITRISE_CERTIFICATE_PASSPHRASE `.

The reason is because we are enabling a new feature en in the platform to detect most common installation issues and as you probably know, most of them are related with the signing certificate and/or provisioning profiles.

Please expect it in a few minutes.

PR #2043, opened 2019-07-08
bluepill-build-test-for-ios 0.2.3

reececomo:
TagCheck


reececomo:
Hey, can anyone see into the black-box for us why all our builds are failing?

Have compared to many other existing steps and made sure the test/audit is air-tight

trapacska:
Hi @reececomo !

The build is failed because the booleans are not as a string in your step.yml:
```
   - type(bool) in value_options with value: false at input: Fail build if coverage fails
   - type(bool) in value_options with value: true at input: Fail build if coverage fails
   - type(bool) in value_options with value: true at input: Only retry tests that failed?
   - type(bool) in value_options with value: false at input: Only retry tests that failed?
```

Please fix these and do a clean-re-share.

Thanks!

PR #2135, opened 2019-09-16
PagerDuty 0.1.0

rwalbergpd:
TagCheck

### DIFF
--- a/steps/pagerduty/0.1.0/step.yml
+++ b/steps/pagerduty/0.1.0/step.yml
@@ -76,7 +76,7 @@ inputs:
     - error
     - warning
     - info
-- skip_if_build_is_passing: true
+- skip_if_build_is_passing: "true"
   opts:
     description: Whether or not this step should be skipped if the build is passing.
       Usually you want this to be set to true.
@@ -85,8 +85,8 @@ inputs:
     summary: Whether or not this step should be skipped if the build is passing
     title: Skip if the build is passing
     value_options:
-    - true
-    - false
+    - "true"
+    - "false"
   
 outputs:
 - PAGERDUTY_EVENT_MESSAGE: null
--- a/steps/pagerduty/0.1.0/step.yml
+++ b/steps/pagerduty/0.1.0/step.yml
@@ -9,7 +9,7 @@ support_url: https://github.com/PagerDuty/bitrise-step-pagerduty/issues
 published_at: 2019-09-16T16:15:01.059024-04:00
 source:
   git: https://github.com/PagerDuty/bitrise-step-pagerduty.git
-  commit: 1b1bf9ab212df4dd745f581d64d6941cf9d84a5b
+  commit: f725ec05df22c26c0926e7599af05e7fc32ffae2
 host_os_tags:
 - osx-10.10
 - ubuntu-16.04
@@ -21,11 +21,10 @@ toolkit:
 deps:
   brew:
   - name: git
-  - name: wget
+  - name: curl
   - name: jq
   apt_get:
   - name: git
-  - name: wget
   - name: curl
   - name: jq
 is_requires_admin_user: true
--- a/steps/pagerduty/0.1.0/step.yml
+++ b/steps/pagerduty/0.1.0/step.yml
@@ -1,0 +1,110 @@
+title: PagerDuty
+summary: |
+  Create a PagerDuty alert
+description: |
+  This step creates a PagerDuty alert with specific parameters. It can use Events API v1 or v2.
+website: https://github.com/PagerDuty/bitrise-step-pagerduty
+source_code_url: https://github.com/PagerDuty/bitrise-step-pagerduty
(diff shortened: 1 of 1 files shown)


rwalbergpd:
I edited `step.yml` by hand before running `$ bitrise share finish`.

Unsure why the build is failing since I don't have access to the details.

steps/pagerduty/0.1.0/step.yml
```diff
@@ -0,0 +1,110 @@
...
+    entry_file: step.sh
+deps:
+  brew:
+  - name: git
+  - name: wget
+  - name: jq
+  apt_get:
+  - name: git
+  - name: wget
+  - name: curl
```
koral--: Curl should be required always, not only on linux since it looks like to be used regardless of OS.

steps/pagerduty/0.1.0/step.yml
```diff
@@ -0,0 +1,110 @@
...
+- ubuntu-16.04
+type_tags:
+- notification
+toolkit:
+  bash:
+    entry_file: step.sh
+deps:
+  brew:
+  - name: git
+  - name: wget
```
koral--: wget does not seem to be used in script

steps/pagerduty/0.1.0/step.yml
rwalbergpd: `curl` is pre-installed on macOS. If you do `$ brew install curl` you will get this message:

```
curl is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.
```

Should I add it anyway?

steps/pagerduty/0.1.0/step.yml
rwalbergpd: Good call. 😅

steps/pagerduty/0.1.0/step.yml
rwalbergpd: Other steps added it so I'll add it too, even if it's not the curl being used.

rwalbergpd:
I made the changes to the `step.yml` in my step repo and in this repo, then bumped the commit in this `step.yml` to match.

No idea why the build is failing since the Details link gives me a 404.

steps/pagerduty/0.1.0/step.yml
```diff
@@ -0,0 +1,109 @@
...
+- skip_if_build_is_passing: true
+  opts:
+    description: Whether or not this step should be skipped if the build is passing.
+      Usually you want this to be set to true.
+    is_expand: false
+    is_required: true
+    summary: Whether or not this step should be skipped if the build is passing
+    title: Skip if the build is passing
+    value_options:
+    - true
```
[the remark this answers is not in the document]
rwalbergpd: Fixed.

I did not make this change in my own step's `step.yml` though. Is that required?

steps/pagerduty/0.1.0/step.yml
```diff
@@ -0,0 +1,109 @@
...
+    is_required: true
+    summary: The description that the event will be created with
+    title: Event description
+    value_options: []
+- severity: error
+  opts:
+    description: The severity that the event will be created with (relevant to Event
+      Rules).
+    is_expand: false
+    is_required: false
```
godrei: Please make this input `is_required: true`, as you have a fixed set of value.

steps/pagerduty/0.1.0/step.yml
```diff
@@ -0,0 +1,109 @@
...
+    summary: The severity that the event will be created with (relevant to Event Rules)
+    title: Event severity (API v2 only)
+    value_options:
+    - critical
+    - error
+    - warning
+    - info
+- skip_if_build_is_passing: "true"
```
godrei: steps have default option `run_if` which can be used to control wether the step should run or not, pls use this option instead of a new step input: https://github.com/bitrise-io/bitrise/blob/master/_examples/experimentals/templates/bitrise.yml#L108

rwalbergpd: @godrei where in the workflow editor can the users of my step control the `run_if`? I want them to decide if it runs on success/failure or only on failure, which is kind of the opposite of most steps, which will run on success, or maybe after a failure.

lszucs: Hi, the `run_if` property can only be set in the `bitrise.yml` mode of the Workflow Editor, there's no option to set it on the UI.

rwalbergpd: 😞 While I agree that exiting functionality should be used over new functionality, running the step only when the workflow is failing is the *default* use case for this step, and having to edit `bitrise.yml` to get the desired behaviour is not something I'd want my company's logo on.

If `run_if` is ever exposed in the UI, I'll re-submit my step, thanks.

rwalbergpd:
@godrei is there a chance we can DM on the Bitrise Slack about this PR? I think some context would go a long way towards understanding why we're doing what we're doing here.

Thanks.

PR #2155, opened 2019-10-01
firebase-app-distribution 0.1.0

guness:
TagCheck

### DIFF
--- a/steps/firebase-app-distribution/0.1.0/step.yml
+++ b/steps/firebase-app-distribution/0.1.0/step.yml
@@ -1,15 +1,15 @@
-title: '[Beta]Firebase App Distribution'
+title: '[BETA] Firebase App Distribution'
 summary: |
-  [Beta]Distributes your builds via Firebase App Distribution
+  [BETA] Distributes your builds via Firebase App Distribution
 description: |
-  [Beta]Uploads your app binaries to Firebase and sends email to your testers.
+  [BETA] Uploads your app binaries to Firebase and sends email to your testers.
 website: https://github.com/guness/bitrise-step-firebase-app-distribution
 source_code_url: https://github.com/guness/bitrise-step-firebase-app-distribution
 support_url: https://github.com/guness/bitrise-step-firebase-app-distribution
-published_at: 2019-10-01T17:14:37.646878+03:00
+published_at: 2019-10-03T15:00:29.831123+03:00
 source:
   git: https://github.com/guness/bitrise-step-firebase-app-distribution.git
-  commit: 0d95b3ad204416fad469d8afd8ebb1ba4848bc86
+  commit: 2f62c82d134b0f1033183050b47f43e4aa6d5981
 project_type_tags:
 - ios
 - android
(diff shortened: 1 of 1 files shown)


steps/firebase-app-distribution/0.1.0/step.yml
```diff
@@ -0,0 +1,108 @@
+title: '[Beta]Firebase App Distribution'
```
trapacska: please have an empty space between the texts and the `[Beta]` part as I am not sure if this way the users can find it by searhing for the firebase keyword.

Also, [BETA] might be more noticeable.

trapacska: apply the comment above for all of this texts

steps/firebase-app-distribution/0.1.0/step.yml
```diff
@@ -0,0 +1,108 @@
...
+      Release notes for this build.
+    is_required: false
+    title: Release Notes
+  release_notes: null
+- opts:
+    description: |
+      Plain text file containing release notes for this build.
+    is_required: false
+    title: Release Notes File
+  release_notes_file: null
```
trapacska: both `_file` input makes not a very real use case on Bitrise. Please remove release_notes_file and testers_file and groups_file.

steps/firebase-app-distribution/0.1.0/step.yml
```diff
@@ -0,0 +1,108 @@
...
+    is_required: false
+    summary: null
+    title: Test Groups
+- groups_file: null
+  opts:
+    description: |
+      Plain text file containing a comma-separated list of group names.
+    is_required: false
+    summary: null
+    title: Test Groups File
```
trapacska: please add an "Additional command flags" input, so the users are able to specify any extra flags they need (also the `_file` suffixed ones)

steps/firebase-app-distribution/0.1.0/step.yml
guness: done

steps/firebase-app-distribution/0.1.0/step.yml
guness: done

steps/firebase-app-distribution/0.1.0/step.yml
guness: done

trapacska:
Thank you for the step @guness ! 🚀 🎉

rohan20:
(I may be asking at the wrong place, please redirect me to the right place otherwise) https://github.com/bitrise-io/bitrise-steplib/pull/2155/files#diff-3e5d14cb744d065472bb2c07a921dce4R14-R19 

Is this supported for Flutter builds too? I don't see it in the list of `project_type_tags`.

guness:
@rohan20 flutter added already. you are looking at first version of the library:
here is the source:

https://github.com/guness/bitrise-step-firebase-app-distribution/blob/master/step.yml

PR #2275, opened 2019-11-22
Appcenter app release 1.0.5

fileformat:
TagCheck

This is the functionality of v1.0.4 that addresses one of the comments with the PR for the 1.0.4 release.  

The step update workflow seems sub-optimal: I can't make a change to the 1.0.4 since it is already tagged in the source repo, so I have to make a new tag (in the source repo) & new PR (in bitrise-steplib).  I also couldn't find the docs for updating a step.  Fortunately, the commands were still in my bash history.

This step will be going away since bitrise is [coming out with their own version](https://github.com/fileformat/bitrise-step-appcenter-app-release/issues/21), so don't think it is worth breaking backwards compatibility on this step, so I have not addressed the pipe vs comma comment.

I also did not address the use of env vars: the script does not use them directly, only in the step.yml, and the snippet is taken directly from the [bitrise docs](https://devcenter.bitrise.io/contributors/create-your-own-step/#secret-environment-variables-in-steps).

### DIFF
--- a/steps/appcenter-app-release/1.0.4/step.yml
+++ b/steps/appcenter-app-release/1.0.4/step.yml
@@ -1,0 +1,84 @@
+title: AppCenter App Release
+summary: |
+  Release an application on Microsoft AppCenter
+description: |
+  Release an application on Microsoft AppCenter
+website: https://github.com/fileformat/bitrise-step-appcenter-app-release
+source_code_url: https://github.com/fileformat/bitrise-step-appcenter-app-release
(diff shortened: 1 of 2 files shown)


lpusok:
@fileformat It seems you included two step.yml-s in the PR, can you have a look?

amarcuse:
The extra was for 1.0.4.  I'm not modifying the repo outside of what stepman is doing.

lpusok:
@amarcuse Can you manually remove the 1.0.4 yml? Sorry for the extra hassle, this is indeed harder than it should be.

PR #2279, opened 2019-11-22
msbuild-sonarscanner-begin 0.9.1

alexshikov:
TagCheck

### Step description

This step allows running SonarScanner for MSBuild against Xamarin projects.
Because of the specifics of the CLI, there are actually two Bitrsie steps developed:
- to run `begin` command
- to run `end` command

Having two Bitrise steps allows using other Bitrise steps in between to run tests, compile, archive, etc. This is the main difference from the already existing Bitrise step developed by **tbrushwyler**: [SonarQube MSBuild Scanner](https://github.com/tbrushwyler/bitrise-step-sonarqube-msbuild-scanner)

- [ x] __I will not move an already shared step version's tag to another commit__
- [x ] I read the [Step Development Guideline](https://github.com/bitrise-io/bitrise/blob/master/_docs/step-development-guideline.md)
- [ x] I have a test for my Step, which can be run with `bitrise run test` (in the step's repository)
- [ x] I did run `bitrise run audit-this-step` (in the step's repository - note, if you don't have this workflow in your `bitrise.yml`, [you can copy it from the step template](https://github.com/bitrise-steplib/step-template/blob/master/bitrise.yml).)
- [ x] I read and accept the [Abandoned Step policy](https://github.com/bitrise-io/bitrise-steplib#abandoned-step-policy)

### DIFF
--- a/steps/msbuild-sonarscanner-begin/0.9.1/step.yml
+++ b/steps/msbuild-sonarscanner-begin/0.9.1/step.yml
@@ -1,0 +1,99 @@
+title: MSBuild SonarScanner (Begin)
+summary: |
+  Step for running SonarScanner agains .NET (Xamarin) projects.
+  This is the first step from two.
+description: "This step runs MSBuild SonarScanner. The scanner for dotnet consists
+  of two steps. \nThis is the first step that executes `begin` CLI command. \nIn between
+  of `begin` and `end` steps the build, test run and code coverage should happen.\n\nCheck
+  official documentation [SonarScanner for MSBuild](https://docs.sonarqube.org/latest/analysis/scan/sonarscanner-for-msbuild/)\nto
+  understand how the scan process happens for dotnet.\n\nThis step can be used for
+  self-hosted SonarQube and SonarCloud.\n"
+website: https://github.com/Diversido/bitrise-step-msbuild-sonarscanner-begin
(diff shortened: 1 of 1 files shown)


trapacska:
Hi @alexshikov !

How this and `-end` step differs from these already shared steps? 
Screen Shot 2019-11-26 at 3 49 09 PM

alexshikov:
@trapacska 
In the existent `sonarqube-scanner` step the begin and end CLI commands wrap a single `msbuild` command where you as a developer may specify a bash command to build Xamarin project. That is useful for a simple build process. But we want to use the whole power of Bitrise and allow to use Bitrise steps in between the begin and end, like:
- Xamarin archive
- NUnit Runner
- etc.

So using these new steps the Bitrise flow can look like this:
- MSBuild Sonar Scanner (Begin)
- NUnit Runner
- Xamarin Archive
- MSBuild Sonar Scanner (End)

alexshikov:
@trapacska does the described reason make sense?

alexshikov:
@trapacska  Here is an example of the same approach of two steps for Azure DevOps:

godrei:
@alexshikov thanks for the new step. The described use-case definitely makes sense.

alexshikov:
@godrei could you please remind, when these steps will be available for use on bitrise.io?

PR #2288, opened 2019-11-26
jira-move-task-step 0.0.1

frakc:
TagCheck

- [*] __I will not move an already shared step version's tag to another commit__
- [*] I read the [Step Development Guideline](https://github.com/bitrise-io/bitrise/blob/master/_docs/step-development-guideline.md)
- [*] I did run `bitrise run audit-this-step` (in the step's repository - note, if you don't have this workflow in your `bitrise.yml`, [you can copy it from the step template](https://github.com/bitrise-steplib/step-template/blob/master/bitrise.yml).)
- [*] I read and accept the [Abandoned Step policy](https://github.com/bitrise-io/bitrise-steplib#abandoned-step-policy)

### DIFF
--- a/steps/jira-move-task-step/0.0.1/step.yml
+++ b/steps/jira-move-task-step/0.0.1/step.yml
@@ -1,0 +1,110 @@
+title: jira-move-task-step
+summary: |
+  This step update optional parameter and assign new status.
+description: |
+  This step update optional parameter and move all tasks from one column to another . Mainly to quckly dump everything from code review to QA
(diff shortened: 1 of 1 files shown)


frakc:
Hello, if the statement below in bold to point that it violates the point, then I want to state that this step does a different thing. A similar step is really published. The difference - published step move only tasks which have their ids in git log; my step move task without consulting with git log. Also in the next update, it will send messages to slack listing links to tasks to ease testers' life. 
I will not move an already shared step version's tag to another commit

PR #2381, opened 2020-02-11
TOOL-1676 Protect step-info files

lpusok:
Git diff is used to check if any step-info file is changed. In this case the step fails, which can be overriden with a retried build. For convenience an curl command is printed to rebuild with the override enabled.


handlers/gate/main.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+		},
+	}
+
+	buildParamsBytes, err := json.Marshal(buildParams)
+	if err != nil {
+		return "", fmt.Errorf("rebuildAPICall: failed to marshal build params, buildParams: %+v, %v", buildParams, err)
+	}
+
+	return fmt.Sprintf(`$ curl https://app.bitrise.io/app/a0bac497f75e1490/build/start.json --data '{"hook_info":{"type":"bitrise","build_trigger_token":" < insert_build_trigger_token > "},
```
lszucs: We might consider factoring the app slug to an env var instead of hard coding it in a public repo.

In case of a future bitrise.io vulnerability, it might be used for malicious purposes.

handlers/gate/step.yml
```diff
@@ -0,0 +1,13 @@
+title: Gate
```
lszucs: I'm not sure if Gate is a well chosen name. It is very general. Based on the description and functionality, it could be  `Reject Files` or something along that line.

handlers/protected-files/main.go
```diff
@@ -12,12 +12,28 @@ import (
...
 type Config struct {
-	BypassGate bool `env:"BYPASS_GATE"`
+	IsApproved bool `env:"changes_approved"`
+}
+
+// BuildTriggerRequestModel ...
+type BuildTriggerRequestModel struct {
+	HookInfo    HookInfoModel    `json:"hook_info"`
+	BuildParams BuildParamsModel `json:"build_params"`
+	UserAgent   string           `json:"triggered_by,omitempty"`
```
lszucs: Although `UserAgent` is completely descriptive, we might consider using the same scheme for mapping properties to json fields.

If json field name is `triggered_by`, this property could be `TriggeredBy` accordingly.

PR #2431, opened 2020-03-23
Tool 1904 validate spec

godrei:
Introducing [JSON schema](https://json-schema.org/understanding-json-schema/reference/index.html) check against the generated StepLib spec.


handlers/validate-spec/main_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test_validateSchema_validSpec(t *testing.T) {
+	require.NoError(t, validateSchema(spec, schema))
+}
+
+func Test_validateSchema_invalidSpec(t *testing.T) {
+	err := validateSchema(invalidSpec, schema)
+	fmt.Println(err)
```
Szabadember: why do we need this?

handlers/validate-spec/main_test.go
```diff
@@ -0,0 +1,65 @@
...
+	require.Error(t, err)
+	require.Equal(t, `/format_version: "1" regexp pattrn ^\d+.\d+.\d+$ mismatch on string: 1
```
Szabadember: string literals could be moved to constants name `expected error`

handlers/validate-spec/main_test.go
```diff
@@ -0,0 +1,65 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_validateSchema_validSpec(t *testing.T) {
```
Szabadember: no underscores

handlers/validate-spec/step.yml
```diff
@@ -0,0 +1,15 @@
...
+website: https://github.com/bitrise-io/bitrise-steplib
+source_code_url: https://github.com/bitrise-io/bitrise-steplib
+support_url: https://github.com/bitrise-io/bitrise-steplib/issues
+
+toolkit:
+  go:
+    package_name: github.com/bitrise-io/bitrise-steplib/handlers/validate-spec
+
+inputs: 
+- spec: $SPEC_JSON
```
trapacska: `spec` -> `spec_path` would describe the expected input format

handlers/validate-spec/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+func main() {
+	var c Config
+	err := stepconf.Parse(&c)
+	if err != nil {
+		failf("Issue with input: %s", err)
+	}
+	stepconf.Print(c)
+
+	schemaPth := filepath.Join(os.Getenv("BITRISE_SOURCE_DIR"), "schema.json")
```
trapacska: this should be an input

handlers/validate-spec/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/go-utils/log"
+	"github.com/bitrise-tools/go-steputils/stepconf"
+	"github.com/qri-io/jsonschema"
+)
+
+// Config ...
+type Config struct {
+	SpecPth string `env:"spec"`
```
trapacska: required input

schema.json
```diff
@@ -0,0 +1,496 @@
...
+        "steplib_source",
+        "download_locations",
+        "assets_download_base_uri",
+        "steps"
+    ],
+    "additionalProperties": false,
+    "properties": {
+        "format_version": {
+            "type": "string",
+            "pattern": "^\\d+.\\d+.\\d+$",
```
trapacska: please test this for `0.0.0` and:
(multi line)
```
1.0.0
1.0.0
```

schema.json
```diff
@@ -0,0 +1,496 @@
...
+            "examples": [
+                "1.0.0"
+            ]
+        },
+        "generated_at_timestamp": {
+            "type": "integer",
```
trapacska: int64 possible?

schema.json
```diff
@@ -0,0 +1,496 @@
...
+            "examples": [
+                1584720375
+            ]
+        },
+        "steplib_source": {
+            "type": "string",
+            "pattern": "^https://github.com/bitrise-io/bitrise-steplib.git$",
```
trapacska: please check all regex for multiline

trapacska: dots are not escaped

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                ],
+                "required": [
+                    "type",
+                    "src"
+                ],
+                "additionalProperties": false,
+                "properties": {
+                    "src": {
+                        "type": "string",
+                        "pattern": "(^https://bitrise-steplib-collection.s3.amazonaws.com/step-archives/$)|(^source/git$)"
```
trapacska: dots are not escaped, please escape dots everywhere

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                    },
+                    "type": {
+                        "type": "string",
+                        "pattern": "(^git$)|(^zip$)"
```
trapacska: two capturing group is needed?
you can have one: `(^git$|^zip$)`
but I think you won't need capturing group at all, so what about: `^git$|^zip$`?

schema.json
trapacska: + same comment as line 63

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                "properties": {
+                                    "icon.svg": {
+                                        "type": "string"
+                                    },
+                                    "icon.png": {
+                                        "type": "string"
+                                    }
+                                }
+                            },
+                            "maintainer": {
```
trapacska: we have a fix set of maintainers currently, pls validate them

handlers/validate-spec/main.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/go-utils/log"
+	"github.com/bitrise-tools/go-steputils/stepconf"
+	"github.com/qri-io/jsonschema"
```
trapacska: I wouldn't allow updating this dependency anymore as we are really expecting a sensitive system to work with this and we cannot afford not noticing a change that has a big impact in our security

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                "type": "string"
+                            }
+                        }
+                    },
+                    "latest_version_number": {
+                        "type": "string",
+                        "pattern": "^\\d+.\\d+.\\d+$",
```
trapacska: dot is not escaped

schema.json
trapacska: dot is not escaped

schema.json
```diff
@@ -0,0 +1,496 @@
...
+            }
+        },
+        "assets_download_base_uri": {
+            "type": "string",
+            "pattern": "^https://bitrise-steplib-collection.s3.amazonaws.com/steps$",
+            "examples": [
+                "https://bitrise-steplib-collection.s3.amazonaws.com/steps"
+            ]
+        },
+        "steps": {
```
trapacska: please add a check for step IDs as they are required also.
- it is not empty
- contains specific set of characters

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                        "type": "string",
+                        "pattern": "^\\d+.\\d+.\\d+$",
+                        "examples": [
+                            "1.0.0"
+                        ]
+                    },
+                    "versions": {
+                        "type": "object",
+                        "propertyNames": {
+                            "pattern": "^\\d+.\\d+.\\d+$"
```
trapacska: same as above

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                            "required": [
+                                "title",
+                                "summary",
+                                "website",
+                                "source_code_url",
+                                "support_url",
+                                "published_at",
+                                "source"
+                            ],
+                            "properties": {
```
trapacska: please validate that the user facing strings contains only an accepted charset

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                "source": {
+                                    "type": "object",
+                                    "required": [
+                                        "git",
+                                        "commit"
+                                    ],
+                                    "additionalProperties": false,
+                                    "properties": {
+                                        "git": {
+                                            "type": "string"
```
trapacska: please validate if ends with `.git`

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                        },
+                                        "commit": {
+                                            "type": "string"
```
trapacska: commit is specific, please check if valid commit string

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                        }
+                                    }
+                                },
+                                "asset_urls": {
+                                    "type": "object",
+                                    "additionalProperties": false,
+                                    "maxProperties": 2,
+                                    "properties": {
+                                        "icon.svg": {
```
trapacska: validate if starts with the only supported base url

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                            "type": "string"
+                                        },
+                                        "icon.png": {
```
trapacska: same as above

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                    "items": {
+                                        "type": "string"
+                                    }
+                                },
+                                "project_type_tags": {
+                                    "type": "array",
+                                    "uniqueItems": true,
+                                    "items": {
+                                        "type": "string",
+                                        "enum": [
```
trapacska: enum doesn't contain empty, however that is also valid, pls check

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                            "utility",
+                                            "dependency",
+                                            "code-sign",
+                                            "build",
+                                            "test",
+                                            "notification"
+                                        ]
+                                    }
+                                },
+                                "dependencies": {
```
trapacska: this is deprecated by now, isn't it?

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                                    "name": {
+                                                        "type": "string"
+                                                    },
+                                                    "bin_name": {
+                                                        "type": "string"
+                                                    }
+                                                }
+                                            }
+                                        },
+                                        "check_only": {
```
trapacska: pls static check for the only accepted value

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                    "type": "boolean"
+                                },
+                                "is_always_run": {
+                                    "type": "boolean"
+                                },
+                                "is_skippable": {
+                                    "type": "boolean"
+                                },
+                                "run_if": {
+                                    "type": "string"
```
trapacska: some kind of format validation? for example can we run the cli's run_if parser on it? (I saw custom validators for this go dep)

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                },
+                                "timeout": {
+                                    "type": "integer"
+                                },
+                                "meta": {
+                                    "type": "object"
+                                },
+                                "inputs": {
```
trapacska: please validate input keys also

schema.json
```diff
@@ -0,0 +1,496 @@
...
+                                                            "type": "string"
+                                                        }
+                                                    },
+                                                    "is_required": {
+                                                        "type": "boolean"
+                                                    },
+                                                    "is_dont_change_value": {
+                                                        "type": "boolean"
+                                                    },
+                                                    "is_template": {
```
trapacska: if this is true then can we check if the value is a valid template?

schema.json
godrei: multi-line handled since the first and last defined (`^..$`)

schema.json
godrei: no, see: https://json-schema.org/understanding-json-schema/reference/numeric.html

schema.json
godrei: multi-line: see above

schema.json
godrei: there is no maintainer field in the spec yet, what is the current list?

schema.json
godrei: What do you mean? What do you suggest to test?

schema.json
godrei: in stepman models, this field has omitempty qualifier, so it will not present in the JSON if has no value and it is not required in the schema: https://github.com/bitrise-io/bitrise-steplib/pull/2431/files#diff-76d5ffeac090c76ed5f4c581f02f3c53R132

the only issue would be if someone defines:

```
project_type_tags:
- ""
```

which i think should not be allowed.

schema.json
godrei: it is deprecated, but in older step versions this field still exists, so it has a format check in the schema

schema.json
godrei: the website's validation is applied: https://github.com/bitrise-io/bitrise-steplib/pull/2431/files#diff-76d5ffeac090c76ed5f4c581f02f3c53R368

schema.json
godrei: no, we can not set conditional checks

schema.json
godrei: i am not sure if we need this, it is like testing if an input in the step.yml is used in the source code.

PR #2467, opened 2020-05-02
version-extractor-android 0.0.1

vladimirpetrovski:
TagCheck

**New Step**

### DIFF
--- a/steps/version-extractor-android/0.0.1/step.yml
+++ b/steps/version-extractor-android/0.0.1/step.yml
@@ -1,0 +1,56 @@
+title: Get versionName and versionCode on Android
+summary: |
+  Extract the versionName and the versionCode on an Android project using a Gradle task
+description: |
+  Extracts the Name version and the Code version from the Android project. It is useful if the versionName and versionCode in `build.gradle` or `build.gradle.kts` are defined somewhere else. Therefore, this step uses a Gradle task to print these two values.
+website: https://github.com/vladimirpetrovski/bitrise-step-version-extractor-android
+source_code_url: https://github.com/vladimirpetrovski/bitrise-step-version-extractor-android
+support_url: https://github.com/vladimirpetrovski/bitrise-step-version-extractor-android/issues
+published_at: 2020-05-02T17:57:15.726671+02:00
+source:
+  git: https://github.com/vladimirpetrovski/bitrise-step-version-extractor-android.git
+  commit: 7e9f33fe3853cf49c8a8880335e37a3ffbfd57ea
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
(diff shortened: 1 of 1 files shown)


steps/version-extractor-android/0.0.1/step.yml
```diff
@@ -0,0 +1,56 @@
...
+- utility
+toolkit:
+  bash:
+    entry_file: step.sh
+is_requires_admin_user: true
+is_always_run: false
+is_skippable: false
+run_if: ""
+inputs:
+- build_gradle_path: $BITRISE_SOURCE_DIR/app/build.gradle
```
koral--, line 28: Step description says about kts files. However, script seems to support only Groovy files.

steps/version-extractor-android/0.0.1/step.yml
```diff
@@ -0,0 +1,56 @@
+title: Get versionName and versionCode on Android
+summary: |
+  Extract the versionName and the versionCode on an Android project using a Gradle task
+description: |
+  Extracts the Name version and the Code version from the Android project. It is useful if the versionName and versionCode in `build.gradle` or `build.gradle.kts` are defined somewhere else. Therefore, this step uses a Gradle task to print these two values.
```
koral--, line 5: Name version -> version name and similarly for code.
It is important that this step reads only values from default config which may be different from actually built variant.

vladimirpetrovski: I'm not sure If I'm getting it right. So, reading the defaultConfig values is right then, isn't it? For example, I'm using `android.`**`defaultConfig`**`.versionName` for the version name which always reads values from default config and not from a particular build variant.

Or maybe I should add an input field where user can define the build variant?

vladimirpetrovski:
@koral-- I've created a new pull request https://github.com/bitrise-io/bitrise-steplib/pull/2469 with some fixes from your comments. I'm closing this one...

steps/version-extractor-android/0.0.1/step.yml
koral--: I meant that if it is not specified in docs which variant is used then it may be unclear for users. Eg. if they are building only one of the variants which uses different values.

Additional input field for variant sounds good. Field may may be optional and default to defualt config if empty.

steps/version-extractor-android/0.0.1/step.yml
vladimirpetrovski: For now, this script supports only Groovy files. If it a requirement from you to support .kts files, please let me know otherwise I'll implement that feature in later versions. Check https://github.com/bitrise-io/bitrise-steplib/pull/2470. Thanks!

PR #2531, opened 2020-06-02
Magic pod 0.54.1

nkns165:
TagCheck

There is already another step named "Magic Pod UI test", but we'd like to add drastic change to input parameters and cannot keep backward compatibility. So we will announce the existing step as deprecated (I will submit another PR).

**New Step**


nkns165:
CI check looks failed.  Can we have any information about the failure?

non-binary:
The issue is that you need to create a step-info.yml file with `maintainer: community`. Here's an example from when I did this in my step: https://github.com/non-binary/bitrise-steplib/commit/cc28eab3890019f55517006d36c232fac2c391b4.

nkns165:
Thank you, updated.  But there is still a problem?

non-binary:
I had to trigger the build from the API (I am a bitrise employee). It now returned green. :)

steps/magic-pod/0.54.1/step.yml
```diff
@@ -0,0 +1,111 @@
...
+website: https://github.com/magic-Pod/bitrise-step-magicpod-uitest
+source_code_url: https://github.com/magic-Pod/bitrise-step-magicpod-uitest
+support_url: https://github.com/magic-Pod/bitrise-step-magicpod-uitest/issues
+published_at: 2020-06-02T22:24:46.286263+09:00
+source:
+  git: https://github.com/Magic-Pod/bitrise-step-magic-pod.git
+  commit: 3503680420460bbd2ab75446c2f6da2767e19df3
+project_type_tags:
+- ios
+- android
```
lpusok, line 15: Do you support multi platform apps like flutter and react-native? Not mandatory but could recommend to more users if such.

nkns165: Thank you for recommendation.  I think our service can handle such multi platform apps, but not especially customized for them.  We will think of such tags in the next version.

PR #2865, opened 2021-01-21
Apps decompiler 1.0.0

imranMnts:
TagCheck

**New Step**


steps/apps-decompiler/1.0.0/step.yml
```diff
@@ -0,0 +1,79 @@
...
+  opts:
+    description: |
+      If you use this step in the same build where you generated the APK
+    is_expand: true
+    is_required: true
+    title: APK path
+- ios_ipa_path: $BITRISE_IPA_PATH
+  opts:
+    description: |
+      If you use this step in the same build where you generated the IPA
```
koral--, line 76: The wording sounds strange. Is a calque?
It's jus an IPA path? What is the point of "If"?

steps/apps-decompiler/1.0.0/step.yml
```diff
@@ -0,0 +1,79 @@
...
+source:
+  git: https://github.com/odemolliens/bitrise-step-apps-decompiler.git
+  commit: 9d55d0cb5e332c4be62952c78329efbf8dd5f8e8
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+project_type_tags:
+- ios
+- android
+- react-native
```
koral--, line 24: What about other multi-platform ones?

steps/apps-decompiler/1.0.0/step.yml
imranMnts: The user will be able to launch this step from the step where he will generate his IPA
In that case, we need to have the path to get it. Per default `XCode archive` step save the path in `$BITRISE_IPA_PATH` but to be able to support the case where the user deployed it in another path we have this input.

In the case where the user launches this step outside of the step where he will generate the IPA (eg. to launch a build specifically for quality to not mix it with deploy build) also we don't need to have this path because it will download the IPA directly from the deployment builds artifacts with `outside_build_slug` input

steps/apps-decompiler/1.0.0/step.yml
imranMnts: Based on 
https://github.com/bitrise-steplib/steps-xcode-archive/blob/master/step.yml
and
https://github.com/bitrise-steplib/steps-gradle-runner/blob/master/step.yml

which generate the APK and IPA
I don't test it with the others so I don't know if it can be used, so to be sure, I didn't add them.

steps/apps-decompiler/1.0.0/step.yml
```diff
@@ -0,0 +1,79 @@
...
+- utility
+toolkit:
+  bash:
+    entry_file: step.sh
+deps:
+  brew:
+  - name: jq
+  - name: apktool
+is_requires_admin_user: false
+is_always_run: true
```
sylank, line 36: Do you want the `is_always_run` field to be `true`? 
```
is_always_run: if true the step will be executed even if a previous step failed during the build. 
Default is false.
- if type_tags is 'notification' this needs to be true
```

sylank (changes requested):

steps/apps-decompiler/1.0.0/step.yml
sylank: I think it is not necessary for this step :)

steps/apps-decompiler/1.0.0/step.yml
koral--: OK, so you should somehow express it in a condensed way for a valid description.

imranMnts:
Continues in #2874

PR #3124, opened 2021-07-19
aws-secrets-manager 1.0.0

Raphx:
TagCheck

### What to do if the build fails?

At the moment contributors do not have access to the CI workflow triggered by StepLib PRs. In case of a failed build, we ask for your patience. Maintainers of Bitrise Steplib will sort it out for you or inform you if any further action is needed.

### New Pull Request Checklist

*Please mark the points which you did / accept.*

- [x] __I will not move an already shared step version's tag to another commit__
- [x] I read the [Step Development Guideline](https://github.com/bitrise-io/bitrise/blob/master/_docs/step-development-guideline.md)
- [x] I have a test for my Step, which can be run with `bitrise run test` (in the step's repository)
- [x] I did run `bitrise run audit-this-step` (in the step's repository - note, if you don't have this workflow in your `bitrise.yml`, [you can copy it from the step template](https://github.com/bitrise-steplib/step-template/blob/master/bitrise.yml).)
- [x] I read and accept the [Abandoned Step policy](https://github.com/bitrise-io/bitrise-steplib#abandoned-step-policy)

PR #3129, opened 2021-07-21
aws-secrets-manager 1.0.0

Raphx:
TagCheck

### DIFF
--- a/steps/aws-secrets-manager/1.0.0/step.yml
+++ b/steps/aws-secrets-manager/1.0.0/step.yml
@@ -1,0 +1,87 @@
+title: Fetch secrets from AWS Secrets Manager
+summary: Fetch secrets from AWS Secrets Manager
+description: |
+  The main intention of this Step is to allow storage of sensitive data on AWS Secrets Manager, and fetching them on-demand during Bitrise workflow execution.
+
+  When a workflow runs, the Step fetches a list of specified secrets from AWS Secrets Manager.
+
+  The fetched secrets are then propagated into subsequent steps in the workflow.
+
+  The Step uses AWS SDK for Go v2 with the default config loader. This means for authenticating with AWS, you may:
+
+    - Use static AWS credentials via environment variable, e.g. `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.
+    - Use shared configuration files, e.g. `AWS_PROFILE`.
+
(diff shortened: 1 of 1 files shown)


Raphx:
Continuation of #3124.

steps/aws-secrets-manager/1.0.0/step.yml
```diff
@@ -0,0 +1,87 @@
...
+  go:
+    package_name: moneylion.com/security/bitrise-step-aws-secrets-manager
+is_requires_admin_user: true
+is_always_run: false
+is_skippable: false
+inputs:
+- aws_access_key_id: ""
+  opts:
+    description: The AWS Access Key ID. If you are supplying this input, ensure the
+      variable is named exactly `AWS_ACCESS_KEY_ID`.
```
godrei, line 35: NOTE: Bitrise Workflow Editor shows the input's title and summary (description replaces summary if you click on the input) by default. This means by default users will not see that the value of this input needs to be `$AWS_ACCESS_KEY_ID`.

It would be better to set the required env vars based on the provided input values. This way users do not need to know what envs are required, they just configure step inputs.

The same applies to `aws_secret_access_key`.

If these inputs are required please add the `is_required: true property`.

godrei (approved):
Thanks for the update, I left one more note, I believe it would improve the step in a further release.

steps/aws-secrets-manager/1.0.0/step.yml
Raphx: Thanks @godrei  for the suggestion. I'm also wondering what's the best way here. If we were to set the required env var, would it be something like:

```yaml
- aws_access_key_id: $AWS_ACCESS_KEY_ID
  opts:
    is_required: true
    is_sensitive: true
```

But this means user would always need to supply `AWS_ACCESS_KEY_ID`, which is not the case, since we want to allow user to use shared configuration via `AWS_PROFILE` as well, without specifying static credentials.

Internally, the Step does not use `aws_access_key_id` nor `aws_secret_access_key` at all. It merely passes the environment variables, which have to be named exactly, to AWS SDK for Go. We kind of rely implicitly on the requirements for AWS SDK for Go here.

In this case, is it possible to set the input as both:
- not required; and
- with a default value for environment variable, i.e. provide a default value for `AWS_ACCESS_KEY_ID`?

PR #3244, opened 2021-10-07
create-a-new-release-in-jira 1.0.0

Emalober:
TagCheck

- [ ] __I will not move an already shared step version's tag to another commit__
- [ ] I read the [Step Development Guideline](https://github.com/bitrise-io/bitrise/blob/master/_docs/step-development-guideline.md)
- [ ] I did run `bitrise run audit-this-step` (in the step's repository - note, if you don't have this workflow in your `bitrise.yml`, [you can copy it from the step template](https://github.com/bitrise-steplib/step-template/blob/master/bitrise.yml).)
- [ ] I read and accept the [Abandoned Step policy](https://github.com/bitrise-io/bitrise-steplib#abandoned-step-policy)

**New Step**

### DIFF
--- a/steps/create-a-new-release-in-jira/1.0.0/step.yml
+++ b/steps/create-a-new-release-in-jira/1.0.0/step.yml
@@ -1,0 +1,100 @@
+title: Create a new release in JIRA
+summary: |
+  Create a new release in JIRA.
+description: |
+  This steps create a Release in your JIRA project, with the start date, publication and description you need.
+website: https://github.com/Emalober/bitrise-step-create-a-new-release-in-jira
+source_code_url: https://github.com/Emalober/bitrise-step-create-a-new-release-in-jira
+support_url: https://github.com/Emalober/bitrise-step-create-a-new-release-in-jira/issues
+published_at: 2021-10-06T21:03:16.937402-03:00
+source:
+  git: https://github.com/Emalober/bitrise-step-create-a-new-release-in-jira.git
+  commit: 567ef82a59926847431c10b3256274ac2ad35bb9
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+type_tags:
+- utility
+toolkit:
+  bash:
+    entry_file: step.sh
+deps:
+  brew:
+  - name: git
+  - name: wget
+  apt_get:
+  - name: git
+  - name: wget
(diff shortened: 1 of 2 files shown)


steps/create-a-new-release-in-jira/1.0.0/step.yml
```diff
@@ -0,0 +1,100 @@
...
+toolkit:
+  bash:
+    entry_file: step.sh
+deps:
+  brew:
+  - name: git
+  - name: wget
+  apt_get:
+  - name: git
+  - name: wget
```
koral--, line 27: You are using curl, but neither git nor wget.

steps/create-a-new-release-in-jira/1.0.0/step.yml
```diff
@@ -0,0 +1,100 @@
...
+    is_required: true
+    summary: Version name
+    title: Version  name
+  version_name: null
+- opts:
+    description: |
+      If the version was published.
+    is_required: true
+    summary: If the version was published
+    title: If the version was published?
```
koral--, line 96: Description of this input does not seem to match the logic behind it.
Past tense here suggests that was already released. Such release event seems to be unrelated to the action performed by the step.
However, in the script you are assigning today's date. So it suggests that the release happens at the time of step execution, not in the past.

You should correct either of that places.

steps/create-a-new-release-in-jira/1.0.0/step.yml
```diff
@@ -0,0 +1,100 @@
...
+    is_required: true
+    summary: Jira company domain
+    title: Jira domain
+- jira_user: null
+  opts:
+    description: |
+      Jira user used to authenticate with Jira API
+
+      e.g. "someone@example.pl"
+    is_sensitive: true
```
koral--, line 59: I don't think that username should be sensitive.
What is a reason behind that?

Emalober: I saw that in many cases, it is used as a sensitive field.

steps/create-a-new-release-in-jira/1.0.0/step.yml
koral--: yes, it should be added.
The only exceptions are bash and go.

PR #3277, opened 2021-11-01
[MOB-222] Add deprecation note to Xamarin steps

tothszabi:
The following Xamarin steps are marked as deprecated:
- nuget-restore
- xamarin-android-test
- xamarin-ios-test
- xamarin-archive
- xamarin-components-restore


shams-ahmed (dismissed):
the only comment I have is should we also add a link? i.e blog post..

ofalvai:
I would also tweak the deprecation notice a bit, a link with more information and the official announcement would be really helpful.

ofalvai (approved):

PR #3351, opened 2022-01-18
scan-with-appsweep 1.0.0

titze:
TagCheck

**New Step**

### DIFF
--- a/steps/scan-with-appsweep/1.0.0/step.yml
+++ b/steps/scan-with-appsweep/1.0.0/step.yml
@@ -1,0 +1,63 @@
+title: AppSweep Mobile Security Testing
+summary: Scan your Android app using Guardsquare's mobile application security testing tool AppSweep
+description: "The build of your app is automatically uploaded to [AppSweep](https://appsweep.guardsquare.com)
+  for scanning. \n\nAppSweep is Guardsquare's free tool enabling developers to put
+  mobile security at the forefront of app development. AppSweep helps you identify
+  and fix security issues in your code and dependencies with actionable recommendations
+  and insights that will help you build more secure mobile apps."
+website: https://www.guardsquare.com/appsweep-mobile-application-security-testing
+source_code_url: https://github.com/Guardsquare/bitrise-step-scan-with-appsweep
+support_url: https://github.com/Guardsquare/bitrise-step-scan-with-appsweep/issues
+published_at: 2022-01-19T13:15:42.006047158+01:00
+source:
+  git: https://github.com/Guardsquare/bitrise-step-scan-with-appsweep.git
+  commit: 74d0913911c69563ddb08b86887fbbce721c1745
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+project_type_tags:
+- android
+type_tags:
+- test
+toolkit:
+  bash:
+    entry_file: step.sh
+is_requires_admin_user: false
+is_always_run: false
+is_skippable: true
+inputs:
+- appsweep_api_key: $APPSWEEP_API_KEY
+  opts:
+    description: |
+      An `APPSWEEP_API_KEY` must be set to allow scanning of the app inside an AppSweep project. You can generate it in the API Keys section of your project settings.
+    is_required: true
+    is_sensitive: true
+    title: AppSweep API key
+- build_variant: release
+  opts:
+    description: |
+      Set to `debug` to upload the debug version of your app, or to `release` to upload the release version.
+    is_required: true
+    title: Select the build variant which should be uploaded
+    value_options:
+    - debug
+    - release
+- opts:
+    description: |
+      Set this to the location of your project inside your repository. Inside this directory, the build file should be `./app/build.gradle` and `gradlew` should be directly in the `project_location`.
+    is_required: true
+    title: Project file path
+  project_location: $PROJECT_LOCATION
+- gradle_plugin_version: 1.0.0
+  opts:
+    description: |
+      The version of AppSweep gradle plugin. Can be set to particular numerical value or to `latest.release`, which requires at least Gradle 7.
+    is_required: true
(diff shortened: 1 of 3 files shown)


steps/scan-with-appsweep/1.0.0/step.yml
```diff
@@ -0,0 +1,60 @@
...
+    category: Config
+    description: |
+      Set this to the location of your project inside your repository.
+    is_required: true
+    title: Project file path
+  project_location: $PROJECT_LOCATION
+- gradlew_path: $GRADLEW_PATH
+  opts:
+    category: Config
+    description: "The Gradle Wrapper is used to start the app scan. \nIf `gradle_path`
```
koral--, line 55:
```suggestion
    description: "The Gradle Wrapper is used to start the app scan. \nIf `gradlew_path`
```

steps/scan-with-appsweep/1.0.0/step.yml
```diff
@@ -0,0 +1,60 @@
+title: AppSweep Mobile Security Testing
+summary: Scan your Android app using Guardsquare's mobile application security testing
+  tool AppSweep
```
ofalvai, line 3: No need to add a line break here, it can be a single line.

steps/scan-with-appsweep/1.0.0/step.yml
```diff
@@ -0,0 +1,60 @@
...
+description: "The build of your app is automatically uploaded to [AppSweep](https://appsweep.guardsquare.com)
+  for scanning. \n\nAppSweep is Guardsquare's free tool enabling developers to put
+  mobile security at the forefront of app development. AppSweep helps you identify
+  and fix security issues in your code and dependencies with actionable recommendations
+  and insights that will help you build more secure mobile apps.\n"
+website: https://www.guardsquare.com/appsweep-mobile-application-security-testing
+source_code_url: https://github.com/Guardsquare/bitrise-step-upload-to-appsweep
```
ofalvai, line 10: This link is 404 for me. Also, it doesn't match the `source` field below.

steps/scan-with-appsweep/1.0.0/step.yml
```diff
@@ -0,0 +1,60 @@
...
+- osx-10.10
+- ubuntu-16.04
+project_type_tags:
+- android
+type_tags:
+- test
+toolkit:
+  bash:
+    entry_file: step.sh
+is_requires_admin_user: true
```
ofalvai, line 26: Are you sure it requires admin user? I see no reason for it.

steps/scan-with-appsweep/1.0.0/step.yml
```diff
@@ -0,0 +1,60 @@
...
+is_always_run: false
+is_skippable: true
+inputs:
+- appsweep_api_key: $APPSWEEP_API_KEY
+  opts:
+    category: Config
```
ofalvai, line 32: If every input is in the category `Config`, then we recommend removing the category altogether, it's not mandatory. The workflow editor will display inputs in the order defined here.

steps/scan-with-appsweep/1.0.0/step.yml
```diff
@@ -0,0 +1,60 @@
...
+    description: |
+      An `APPSWEEP_API_KEY` must be set to allow scanning of the app inside an AppSweep project. You can generate it in the API Keys section of your project settings.
+    is_required: true
+    is_sensitive: true
+    title: AppSweep API key
+- debug: false
```
ofalvai, line 38: A boolean input is usually modeled as a dropdown with two values in Bitrise steps. Please use the `value_options` field, like here: https://github.com/bitrise-steplib/steps-xcode-archive/blob/master/step.yml#L369

steps/scan-with-appsweep/1.0.0/step.yml
```diff
@@ -0,0 +1,60 @@
...
+  opts:
+    category: Config
+    description: |
+      Set to true to upload the debug version of your app, or to false (default) to upload the release version.
+    is_required: true
+    title: Should debug version be uploaded
+- opts:
+    category: Config
+    description: |
+      Set this to the location of your project inside your repository.
```
ofalvai, line 48: It's not clear from the title and description what this path should point to. Is it the path to the `build.gradle` file, or just the folder where the `build.gradle` file is?

ofalvai (changes requested):
Thank you @titze for the contribution. I left some comments on the `step.yml` file.

I also had a look at the step source code, and have a few observations:

1. I see that `$APPSWEEP_UPLOAD_URL` is exposed at the end of the script, but this env var is missing from the `outputs` section of `step.yml`. Please declare this env var as an output so that the workflow editor can display it for users. Also, as a best practice, we recommend steps to print a log message about the output name and its actual value (if it's not sensitive) for discoverability.
2. I see that the injected Gradle plugin's version is harcoded into the shell script. In our experience with steps, it can cause unexpected problems and hard to debug user issues. We recommend creating a step input for this version number, where you can add a default value. It makes it easier to update the version number in the future too.

steps/scan-with-appsweep/1.0.0/step.yml
titze: This indeed was the wrong link.

Does it need to be the same as the Website? The website would explain what AppSweep is, the source code is for this step.

steps/scan-with-appsweep/1.0.0/step.yml
titze: this is a bit weird, in our repo this is on one line, but when creating this in the steplip using `bitrise share` it adds this line break.

ofalvai: Oh, then please ignore it, we'll investigate this.

ofalvai (approved):
Thank you @titze for making the requested changes. It's perfect now.

PR #3605, opened 2022-09-07
Flutter version auto increment 0.0.1

IbrahimTabba:
TagCheck

**New Step**

### DIFF
--- a/steps/flutter-version-auto-increment/0.0.1/step.yml
+++ b/steps/flutter-version-auto-increment/0.0.1/step.yml
@@ -1,0 +1,67 @@
+title: Flutter version auto increment
+summary: |
+  Increment current version and build number in pubspec.yaml file.
+description: |
+  Increment Project current version and build number file and rewite new value in pubspec.yaml file.
+website: https://github.com/IbrahimTabba/bitrise-step-flutter-version-auto-increment
+source_code_url: https://github.com/IbrahimTabba/bitrise-step-flutter-version-auto-increment
+support_url: https://github.com/IbrahimTabba/bitrise-step-flutter-version-auto-increment/issues
+published_at: 2022-09-07T23:50:57.142685+03:00
+source:
+  git: https://github.com/IbrahimTabba/bitrise-step-flutter-version-auto-increment.git
+  commit: ed46101375385f8cc36e254da03b574814d294da
+project_type_tags:
+- flutter
+type_tags:
+- deploy
+- build
+toolkit:
+  bash:
+    entry_file: step.sh
+inputs:
+- opts:
+    is_expand: true
+    is_required: true
+    title: Path to pubspec.yaml
+  pubspec_path: $BITRISE_SOURCE_DIR/pubspec.yaml
+- opts:
+    description: |
+      patch: X.X.0 -> X.X.1
+      minor: X.0.X -> X.1.0
+      major: 1.X.X -> 2.0.0
+    is_expand: true
+    is_required: true
+    title: Version auto increment mode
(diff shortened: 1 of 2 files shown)


steps/flutter-version-auto-increment/0.0.1/step.yml
```diff
@@ -0,0 +1,68 @@
+title: Flutter version auto increment
+summary: |
+  Increment current version and build number in pubspec.yaml file.
+description: |
+  Increment Project current version and build number file and rewite new value in pubspec.yaml file.
```
lpusok, line 5:
```suggestion
  Increment Project current version and build number file and write new value in pubspec.yaml file.
```

steps/flutter-version-auto-increment/0.0.1/step.yml
```diff
@@ -0,0 +1,68 @@
...
+    is_expand: true
+    is_required: true
+    title: Version auto increment mode
+    value_options:
+    - patch
+    - minor
+    - major
+    - none
+  version_auto_increment_mode: patch
+- increment_build_number: "yes"
```
lpusok, line 41: Suggestion: To be consistent with iOS/Android versioning, could use Bitrise build number + a configurable offset (see https://devcenter.bitrise.io/en/builds/build-numbering-and-app-versioning.html#offsetting-the-build-version).

lpusok (changes requested):
Thanks for the contribution.
Please update the title and the type in the description, then re-share the Step.

steps/flutter-version-auto-increment/0.0.1/step.yml
```diff
@@ -0,0 +1,68 @@
+title: Flutter version auto increment
```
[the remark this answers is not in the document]
koral--: Both titles suggest that the subject is a version of Flutter itself, not a version of the project inside.

steps/flutter-version-auto-increment/0.0.1/step.yml
IbrahimTabba: the source for both version and build number in flutter apps is the "version: X.X.X+X" stored in pubspec.yaml file.
this step is meant to be for production workflows, where the build number and version should be exactly one increment from the previous build, regardless of how many builds I did on bitrise and this variable BITRISE_BUILD_NUMBER

lpusok: Based on the [pubspec docs](https://dart.dev/tools/pub/pubspec#version) and [semver](https://semver.org/#spec-item-10), the build metadata (denoted by `+`) can be almost any string (git hash prefix for example) and it is not considered when comparing two versions.
My question is can we be sure that there is always a build number metadata commited? Incrementing a version without metadata (`1.2.3`) would not work. The same questoin also for a prerelease version being commited (`1.2.3-alpha`).

For reference checked an example package to understand conventions: https://pub.dev/packages/webview_flutter/versions.
Based on these, it seems that build number is usually not added, only for second build with the same minor, major and patch versions. Sometimes a prerelease version is used (`-alpha1` for example)

lpusok: An other idea: is this the input `increment_build_number` needed? It seems to me that `version_auto_increment_mode` option `none` could mean that we only increase build number. (Or add it if it was not yet added). 
The `none` option could even be renamed to `build_number_only` or similar.

lpusok (approved):

lpusok:
@IbrahimTabba Should we release as is?

PR #3721, opened 2023-01-25
maestro-cloud-upload 1.1.0

ArthurSav:
TagCheck

### DIFF
--- a/steps/maestro-cloud-upload/1.1.0/step.yml
+++ b/steps/maestro-cloud-upload/1.1.0/step.yml
@@ -1,0 +1,177 @@
+title: Upload to Maestro Cloud
+summary: |
+  Upload your app binary to Maestro Cloud and run your flows
+description: |
+  This steps uploads both your app binary and flows to Maestro Cloud.
+website: https://github.com/mobile-dev-inc/bitrise-step-maestro-cloud-upload
+source_code_url: https://github.com/mobile-dev-inc/bitrise-step-maestro-cloud-upload
+support_url: https://github.com/mobile-dev-inc/bitrise-step-maestro-cloud-upload/issues
+published_at: 2023-01-25T22:06:55.504814+04:00
+source:
+  git: https://github.com/mobile-dev-inc/bitrise-step-maestro-cloud-upload.git
+  commit: e418f3fa9c658168265c971e3064d95bea7e6cff
+host_os_tags:
+- osx-10.10
+- ubuntu-16.04
+project_type_tags:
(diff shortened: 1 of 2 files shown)


steps/maestro-cloud-upload/1.1.0/step.yml
```diff
@@ -0,0 +1,177 @@
...
+    is_expand: true
+    is_required: false
+    summary: Pass environment variables to your flows
+    title: Environment variables
+- android_api_level: ""
+  opts:
+    description: ""
+    is_expand: true
+    is_required: false
+    summary: 'Set the Android api level the devices should run (Default: 30)'
```
ofalvai, line 100: If this has a default value, I recommend defining it explicitly in line 95. This way it will show up in the workflow editor as the current value of the input.

steps/maestro-cloud-upload/1.1.0/step.yml
```diff
@@ -0,0 +1,177 @@
...
+    title: Include tags
+- exclude_tags: ""
+  opts:
+    description: ""
+    is_expand: true
+    is_required: false
+    summary: Exclude flows from running with the specified tags (comma separated)
+      i.e pull-request,experimental
+    title: Exclude tags
+- export_format: "false"
```
ofalvai, line 118: The name of this input seems to be different than what the `title` suggests. I guess this should be `export_test_report`, right?

ofalvai (changes requested):
Just a couple of minor issues with the inputs, looking forward to merge this!

steps/maestro-cloud-upload/1.1.0/step.yml
ArthurSav: Hi @ofalvai the reason i left the default value empty is because it can also be an iOS app, in which case it doesn't make sense to send the android api level unless explicitly set.

ofalvai: Gotcha, makes sense!

steps/maestro-cloud-upload/1.1.0/step.yml
ArthurSav: Good observation. It was meant for export format, but we only support `junit` right now so i made it a boolean.
I will update it

ArthurSav:
@ofalvai Can you check again, i believe i addressed the comments

ofalvai (approved):
Looks good, thank you for the fixes @ArthurSav!

PR #3835, opened 2023-05-21
appdome-build-2secure-android 1.0.7

avi112211:
TagCheck

Update to existing step:

added return values


godrei (approved):
Hi @avi112211 ,
sorry for the late response, the new step version looks good to me!

PR #3865, opened 2023-06-21
appdome-build-2secure-android 1.0.11

elish266:
TagCheck

change google sign to use fingerprint if google_fingerprint is not provided.


tothszabi:
The commit hash in the yaml file for the release 1.0.11 is different then the hash in your repo. The two has to match so could you fix that?

avi112211:
Thanks @tothszabi 
new PR: https://github.com/bitrise-io/bitrise-steplib/pull/3870

PR #4077, opened 2024-01-09
firebase-test-labs 0.0.2

AnthonyDadeWT:
TagCheck

- [ x ] __I will not move an already shared step version's tag to another commit__
- [ x ] I read the [Step Development Guideline](https://github.com/bitrise-io/bitrise/blob/master/_docs/step-development-guideline.md)
- [ x ] I have a test for my Step, which can be run with `bitrise run test` (in the step's repository)
- [ x ] I did run `bitrise run audit-this-step` (in the step's repository - note, if you don't have this workflow in your `bitrise.yml`, [you can copy it from the step template](https://github.com/bitrise-steplib/step-template/blob/master/bitrise.yml).)
- [ x ] I read and accept the [Abandoned Step policy](https://github.com/bitrise-io/bitrise-steplib#abandoned-step-policy)

**New Step**

### DIFF
--- a/steps/firebase-test-labs/0.0.2/step.yml
+++ b/steps/firebase-test-labs/0.0.2/step.yml
@@ -1,0 +1,234 @@
+title: '[Beta] Firebase Test Labs'
+summary: "[Beta] Automatic deployment of iOS and Android Flutter Integration tests
+  through Firebase Test Labs \n"
+description: "[Beta] Automatic deployment of iOS and Android Flutter Integration tests
+  to Firebase Test Labs \n\nThis step allows you to automate the process of deploying
+  tests to be run on **physical devices** on Firebase Test Labs.\n\n- Before using
+  this step, ensure to go through the setup required to [test your app on virtual
+  devices](https://github.com/flutter/flutter/tree/main/packages/integration_test#android-device-testing)\n\n-
+  **NOTE**: To run your project's tests on an iOS device on Firebase Test Labs you
+  must properly \ncodesign your app in the workflow before this step is run. [The
+  documentation can be found here](https://devcenter.bitrise.io/en/code-signing/ios-code-signing/managing-ios-code-signing-files---automatic-provisioning.html)\n\n-
+  If you would not like to follow the documentation's guidance on using automatic
+  code signing, you could also choose\nto upload a proper provisioning profile for
+  your iOS application to the Code Signing & Files tab in the workflow editor\non
+  Bitrise. Then, add the \"Certificate and profile installer\" step (without editing
+  the configuration) to your \nworkflow before this step is run.\n\n- To save build
+  credits and time, this step runs the firebase test command with the --async flag,
+  so the command can \nbe run without waiting for test results to come back. This
+  is not to be confused with the timeout input below."
+website: https://github.com/AnthonyDadeWT/bitrise-step-firebase-test-labs
+source_code_url: https://github.com/AnthonyDadeWT/bitrise-step-firebase-test-labs
+support_url: https://github.com/AnthonyDadeWT/bitrise-step-firebase-test-labs/issues
+published_at: 2024-01-09T15:45:44.626338-05:00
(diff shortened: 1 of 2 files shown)


AnthonyDadeWT:
Let me know of any questions, suggestions, or action items I need to complete. Thank you!

ofalvai (changes requested):
Hello @AnthonyDadeWT 👋 

A general observation before doing a deeper review of this step: it's probably not a good idea to do both the Flutter build and the Firebase testing in a single step. There are existing Flutter steps for building iOS or Android artifacts, this step should probably work with the [outputs of the build step](https://github.com/bitrise-steplib/bitrise-step-flutter-build/blob/master/step.yml#L182).

It's easier to debug build issues when one step is doing one thing only. Also, this step could be useful for non-Flutter projects too, if it was working with standard app artifacts.

AnthonyDadeWT:
@ofalvai 

Hello, and thank you for your review!

> A general observation before doing a deeper review of this step: it's probably not a good idea to do both the Flutter build and the Firebase testing in a single step. There are existing Flutter steps for building iOS or Android artifacts, this step should probably work with the [outputs of the build step](https://github.com/bitrise-steplib/bitrise-step-flutter-build/blob/master/step.yml#L182)

This is a good point. I did implement this for Android, on line 101 of my step.sh starts the logic of checking whether or not we need to build an APK based on if $BITRISE_APK_PATH exists. However I don't think this would actually be possible on the iOS side, [because according to flutter's documentation on deploying integration tests to FTL ](https://github.com/flutter/flutter/tree/main/packages/integration_test#ios-device-testing), you must first run `flutter build ios $integration_test_path --release` for this process to actually work. So what do you think about this case? Maybe a "Flutter Build for Testing" Step should be created in conjunction with this step to use together like you said. 

> It's easier to debug build issues when one step is doing one thing only. Also, this step could be useful for non-Flutter projects too, if it was working with standard app artifacts.

This is a thought I had as well since this Step should be compatible. However, I was afraid that since Bitrise has "Device Testing" Steps for iOS and Android already, I feared that it would be considered as a duplicate step. Although I will say there are major differences from the Device Testing Steps and the one I've created here, where this Step uses the user's Firebase project, runs the tests asynchronously, and the results are displayed there on Firebase instead of Bitrise.

ofalvai:
I see, thank you for the extra context and details :) In this case, I'm thinking maybe the best compromise would be to add "Flutter" to the step title to avoid confusion and differentiate it from other Firebase steps. What do you think about this compromise?

AnthonyDadeWT:
Yes that makes sense to me. I thought I couldn't have a platform's name in the title due to the rules on Steps but I see other Steps have Platforms in the name.

Here is my plan for updating this step to follow your guidance.

Change to accept outputs from

Flutter Build (for a release build IPA)
Android Build for UI Testing (for an APK and TestAPK)
Xcode Build for testing (.xctestrun)

Then modify the Step to accept these outputs and only run the execution of sending these items to FTL.

Update documentation to guide user to use these Steps before this Step in a workflow.

Change name to "Flutter Device Testing" ? - open to suggestions here

Let me know how that all sounds. It will take me some time to get this done.

ofalvai:
@AnthonyDadeWT the plan looks good to me, thanks for outlining this!

Naming: I would keep "Firebase" in the title because of your previous point that it does something similar to the first-party step, but with the user's Firebase account. So anyone who wants to set up Firebase testing in a custom way would find your step.

godrei:
Hi @AnthonyDadeWT I'm closing this PR for now, because, as discussed, some changes are required for this Step.
Here is a guide on how to fix a StepLib PR issue: https://devcenter.bitrise.io/en/steps-and-workflows/developing-your-own-bitrise-step/sharing-steps-with-all-bitrise-users.html#fixing-issues-in-a-steplib-pull-request
Please link this PR, when you open a new one with the updated Step, to let us have the necessary context.

PR #4419, opened 2025-03-28
bitrise-step-install-bundler: remove duplicate step

ofalvai:
TagCheck

@igorcferreira I see the repo is now archived, so I couldn't open an issue there. Our analytics don't show any usage of the step ID `bitrise-step-install-bundler`, so I think it's safe to clean up, but please let me know if you have concerns.


igorcferreira:
Hello @ofalvai. Yes, as you mentioned, the step `install-bundler` is archived. It is perfectly fine to clean-up

tothszabi (approved):

